PR #5, opened 2022-08-24
Base repo

justkawal:
- Clean base repo
- Does not contain any hand modified code


justkawal:
It's much better to use polkadart_scale_codec branch for developing that respective package only.

leonardocustodio (approved):

Lohann (approved):

PR #9, opened 2022-08-25
SS58 Codec Implementation.

justkawal:
- **SS58Codec.decode** and **SS58Codec.encode**
- Test Cases to encode and decode back and forth.
- SS58Codec will provide encode and decode functionality to **ss58**.

### DIFF
--- a/packages/ss58_codec/lib/src/ss58_codec_base.dart
+++ b/packages/ss58_codec/lib/src/ss58_codec_base.dart
@@ -4,38 +4,56 @@ import 'dart:typed_data';
 import 'package:base_x/base_x.dart';
 import 'package:cryptography/dart.dart';
 import 'package:ss58_codec/src/address.dart';
+import 'package:ss58_codec/src/exceptions.dart';
 
-// Hash Prefix to be added before hashing data
+/// [Private]
+///
+/// Hash Prefix to be added before hashing data.
 final List<int> _hashPrefix = utf8.encode('SS58PRE');
 
-// base58 hasher
+/// [Private]
+///
+/// Fast base encoding / decoding of any given alphabet using bitcoin style leading zero compression.
 final _base58 =
     BaseXCodec('123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz');
 
+/// Key that can be encoded to/from SS58.
+///
+/// For more information see [Substrate (ss58)](https://docs.substrate.io/v3/advanced/ss58/)
 class SS58Codec {
   ///
   /// Decode SS58 address string.
   ///
   /// This function follows [Ss58Codec trait](https://github.com/paritytech/substrate/blob/ded44948e2d5a398abcb4e342b0513cb690961bb/primitives/core/src/crypto.rs#L245)
-  static Address decode(String s) {
-    var data = _base58.decode(s);
+  ///
+  /// ```
+  /// [Exceptions]
+  /// throw InvalidPrefixException:    when (first byte of decoded address > 127)
+  /// throw BadAddressLengthException: when (address.length < 3)
+  /// throw InvalidCheckSumException: when hashed data is invalid
+  /// ```
+  static Address decode(String address) {
+    var data = _base58.decode(address);
     if (data.length < 3) {
-      throw invalidAddress(s);
+      throw BadAddressLengthException(address);
     }
-    var data0 = data[0];
     late int offset;
     late int prefix;
-    if (data0 < 64) {
-      prefix = data0;
+    if (data[0] < 64) {
+      prefix = data[0];
       offset = 1;
-    } else if (data0 < 128) {
-      var b1 = data[1];
-      var lower = ((data0 << 2) | (b1 >> 6)) & BigInt.from(255).toInt();
-      var upper = b1 & BigInt.from(63).toInt();
-      prefix = lower | (upper << 8);
+    } else if (data[0] < 128) {
+      // weird bit manipulation owing to the combination of LE encoding and missing two
+      // bits from the left.
+      // d[0] d[1] are: 01aaaaaa bbcccccc
+      // they make the LE-encoded 16-bit value: aaaaaabb 00cccccc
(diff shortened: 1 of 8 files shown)


packages/ss58_codec/test/ss58_codec_test.dart
```diff
@@ -1,16 +1,94 @@
...
+    /// hydradx
+    decodeTest('7LKANFqnxGZ5zVyQNsiNZM2n9dSVLNzmkwKoUKR7AALea3vN', 63);
+
+    /// crust
+    decodeTest('cTMxUeDi2HdYVpedqu5AFMtyDcn4djbBfCKiPDds6k1fuFYXL', 66);
+
+    /// subspace testnet
+    decodeTest('st6v8nztLTbiqY5Hw97L5FWCBmpzMsrAh5qXZ1tJs1epNvoFA', 2254);
+
+    /// basilisk
```
leonardocustodio, line 43: You only test the happy path.
We should test the exceptions too. Like having an invalid prefix, or the address being the wrong size, etc.

leonardocustodio (changes requested):

packages/ss58_codec/test/ss58_codec_test.dart
justkawal: Done

leonardocustodio (dismissed):

packages/ss58_codec/lib/src/ss58_codec_base.dart
```diff
@@ -1,4 +1,138 @@
...
+      prefix: prefix,
+      bytes: data.sublist(offset, data.length - hashLen),
+    );
+  }
+
+  ///
+  /// Encode SS58 address into canonical string format
+  static String encode(Address address) {
+    final int prefix = address.prefix;
+    assert(prefix >= 0 && prefix < 16384, 'invalid prefix');
```
Lohann, line 74: I don't think we should use assert here, asserts are only verified in `debug` mode, not in `release` mode, more info here:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use-asserts-liberally-to-detect-contract-violations-and-verify-invariants

> assert() allows us to be diligent about correctness without paying a performance penalty in release mode, because Dart only evaluates asserts in debug mode.

Instead, you should create and throw custom Exceptions, example:
https://github.com/flutter/flutter/blob/3.3/packages/flutter_tools/lib/src/build_system/exceptions.dart#L9-L37

packages/ss58_codec/lib/src/ss58_codec_base.dart
```diff
@@ -1,4 +1,138 @@
...
+    }
+
+    return Address(
+      prefix: prefix,
+      bytes: data.sublist(offset, data.length - hashLen),
+    );
+  }
+
+  ///
+  /// Encode SS58 address into canonical string format
```
Lohann, line 71: This description also should include what exceptions this method throws, and what pre-conditions are verified

packages/ss58_codec/lib/src/ss58_codec_base.dart
```diff
@@ -1,4 +1,138 @@
...
+    data.setAll(offset, address.bytes);
+    final List<int> hashedData = computeHash(data, hashLen);
+    for (var i = 0; i < hashLen; i++) {
+      data[offset + len + i] = hashedData[i];
+    }
+    return _base58.encode(data);
+  }
+}
+
+List<int> computeHash(Uint8List data, int length) {
```
Lohann, line 119: Documentation missing for this method

packages/ss58_codec/lib/src/ss58_codec_base.dart
```diff
@@ -1,4 +1,138 @@
...
+import 'package:ss58_codec/src/address.dart';
+
+// Hash Prefix to be added before hashing data
+final List<int> _hashPrefix = utf8.encode('SS58PRE');
+
+// base58 hasher
+final _base58 =
+    BaseXCodec('123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz');
+
+class SS58Codec {
```
Lohann, line 15: Documentation missing for this class

packages/ss58_codec/lib/src/ss58_codec_base.dart
```diff
@@ -1,4 +1,138 @@
...
+  ///
+  /// Decode SS58 address string.
+  ///
+  /// This function follows [Ss58Codec trait](https://github.com/paritytech/substrate/blob/ded44948e2d5a398abcb4e342b0513cb690961bb/primitives/core/src/crypto.rs#L245)
+  static Address decode(String s) {
```
Lohann, line 20: I think more granular exceptions are needed here too, `invalidAddress` is too generic. 

Example of Exceptions:
https://github.com/paritytech/substrate/blob/polkadot-v0.9.27/primitives/core/src/crypto.rs#L200-L230

Example on how SS58 is decoded in rust:
https://github.com/paritytech/substrate/blob/polkadot-v0.9.27/primitives/core/src/crypto.rs#L262-L304

Lohann (changes requested):

packages/ss58_codec/lib/src/ss58_codec_base.dart
justkawal: Done ✅

packages/ss58_codec/lib/src/ss58_codec_base.dart
justkawal: Done ✅

packages/ss58_codec/lib/src/ss58_codec_base.dart
justkawal: Done ✅

packages/ss58_codec/lib/src/ss58_codec_base.dart
justkawal: Rust Codes are pretty well documented and makes it much easier to understand things now-a-days.

Thanks.

Lohann (approved):
Approved!

PR #20, opened 2022-09-02
Feature/CI-CD

justkawal:
CI-CD
- Check Lint
- Check dprint formatting
- Check Formatting
- Analyse Dart Code
- Dart Code Testing

Note:
Used **reusable workflow** to check for following tests/commands in each package and in polkadart package.
- Dart check formatting
- Dart analyze
- Dart test


.github/workflows/reuseable_workflow.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ inputs.working-directory }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
```
Lohann, line 19: I don't think is necessary to run check out code here again (this basically clones and checkout to the right branch), once it already runs in the parent workflow.

justkawal: It throws error when removed

Lohann (approved):
approved!

PR #21, opened 2022-09-02
Update reuseable_workflow.yml

leonardocustodio:


justkawal:
The code coverage will work by using the latest reuseable-workflow and specifying the current branch name after @ in test.yml in place of feature/CI-CD

leonardocustodio:
Feel free to update ;)

leonardocustodio:
Hmmm I believe @Lohann actions are over :laughing:

leonardocustodio:
That seems wrong, before we had different percentages for each project. Now all of them are the same. I believe it is getting the same coverage file for all of them.

justkawal:
> That seems wrong, before we had different percentages for each project. Now all of them are the same. I believe it is getting the same coverage file for all of them.

Yes, I looked into it, I was trying to optimise the speed by making tests all in single run and formatting in single run, but it took 1 minute more than before.

Lohann (approved):

Issue #23, opened 2022-09-08
Implementation of code generator using annotations


LuizaJurimetric:
1/4 related to #13 

## Description 
Implementation of scale-codec decorators with dart code generator and annotations. The main concept is about generating classes with the right scale codec methods.

## Acceptance Criteria
- must create a new .g.dart file with the class extension when user writes the decorator above class name.

gabrielokura:
Will be merged on milestone 2

Issue #26, opened 2022-09-08
Implementation of class decoding


LuizaJurimetric:
4/4 related to #13 

## Description 
Implementation of the method 'decode' on any class by creating a new dart file and without writing this method on each one 

## Acceptance Criteria
- must create a new .g.dart file with the class extension, including decode function.

Issue #25, opened 2022-09-08
Implementation of class encoding


LuizaJurimetric:
3/4 related to #13 

## Description 
Implementation of the method 'encode' on any class by creating a new dart file and without writing this method on each one 

## Acceptance Criteria
- must create a new .g.dart file with the class extension, including encode function.

PR #27, opened 2022-09-14
Feature/code generator (1/4)

gabrielokura:
## PR of the issue #23 
- [x] Scale Codec annotation package
- [x] Scale Codec code generator using @ScaleCodecSerializable
- [x] Class generation tests
- [x] Encoded fields tests


packages/scale_codec_serializable/lib/src/decoder_helper.dart
```diff
@@ -0,0 +1,41 @@
...
+    yield buffer.toString();
+  }
+
+  void _writeDefaultConstructor(
+      StringBuffer stringBuffer, Map<String, FieldElement> fields) {
+    stringBuffer.write('$targetClassReference(');
+
+    for (var field in fields.keys) {
+      final fieldName =
+          field.startsWith('_') ? field.replaceFirst('_', '') : field;
```
pastre, line 34: This check is not necessary, `replaceFirst` will only replace the first character if the string begins with said character

packages/scale_codec_serializable/lib/src/encoder_helper.dart
```diff
@@ -0,0 +1,23 @@
...
+  /// Write encode method implementation as example above:
+  ///
+  /// ```dart
+  ///   // ...
+  ///   String encode() => "";
+  ///   // ...
+  /// ```
+  Iterable<String> createEncode(
+      Map<String, FieldElement> accessibleFields) sync* {
+    assert(config.createDecodeMethod);
```
pastre, line 14: Same as above

packages/scale_codec_serializable/lib/src/generator_helper.dart
```diff
@@ -0,0 +1,93 @@
...
+          config,
+        );
+
+  @override
+  void addMember(String memberContent) {
+    _addedMembers.add(memberContent);
+  }
+
+  Iterable<String> generate() sync* {
+    assert(_addedMembers.isEmpty);
```
pastre, line 26: Same as above

packages/scale_codec_serializable/lib/src/generator_helper.dart
```diff
@@ -0,0 +1,93 @@
+import 'package:analyzer/dart/element/element.dart';
+import 'package:build/build.dart';
+import 'package:scale_codec_serializable/src/class_visitor.dart';
+import 'package:scale_codec_serializable/src/decoder_helper.dart';
+import 'package:scale_codec_serializable/src/encoder_helper.dart';
+import 'package:scale_codec_serializable/src/helper_core.dart';
+import 'package:scale_codec_serializable/src/type_helpers/config_types.dart';
+import 'package:source_gen/source_gen.dart';
+
+class GeneratorHelper extends HelperCore with EncoderHelper, DecodeHelper {
```
pastre, line 10: This is missing tests

packages/scale_codec_serializable/lib/src/generator_helper.dart
```diff
@@ -0,0 +1,93 @@
...
+  ///   bool isTest = true;
+  ///   String get isTestEncoded => "0x01";
+  ///   //...
+  /// ```
+  String createClassBody(Map<String, FieldElement> fields) {
+    final buffer = StringBuffer();
+
+    for (var field in fields.keys) {
+      final fieldName =
+          field.startsWith('_') ? field.replaceFirst('_', '') : field;
```
pastre, line 84: Same as above

packages/scale_codec_serializable/lib/src/helper_core.dart
```diff
@@ -0,0 +1,47 @@
...
+import 'package:analyzer/dart/element/type.dart';
+import 'package:meta/meta.dart';
+import 'package:source_gen/source_gen.dart';
+import 'package:source_helper/source_helper.dart';
+
+import 'type_helpers/config_types.dart';
+import 'unsupported_type_error.dart';
+import 'utils.dart';
+
+abstract class HelperCore {
```
pastre, line 11: I didn't understand: why are we creating this abstraction? Good practice is to use composition over inheritance.
Also missing tests

packages/scale_codec_serializable/lib/src/scale_codec_serializable_generator.dart
```diff
@@ -0,0 +1,37 @@
...
+
+class ScaleCodecSerializableGenerator
+    extends GeneratorForAnnotation<ScaleCodecSerializable> {
+  final ClassConfig _config;
+  ScaleCodecSerializableGenerator.withConfig(this._config);
+
+  @override
+  Iterable<String> generateForAnnotatedElement(
+      Element element, ConstantReader annotation, BuildStep buildStep) {
+    if (!element.library!.isNonNullableByDefault) {
```
pastre, line 17: We shouldn't force unwrap

packages/scale_codec_serializable/lib/src/scale_codec_serializable_generator.dart
```diff
@@ -0,0 +1,37 @@
...
+      throw InvalidGenerationSourceError(
+          'Generator cannot target libraries that have not been migrated to '
+          'null-safety.',
+          element: element);
+    }
+
+    if (element is! ClassElement || element is EnumElement) {
```
pastre, line 24: Would this crash if `element` is not of type `ClassElement`?

packages/scale_codec_serializable/lib/src/scale_part_builder.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:scale_codec_annotation/scale_codec_annotation.dart';
+import 'package:scale_codec_serializable/src/scale_codec_serializable_generator.dart';
+import 'package:scale_codec_serializable/src/type_helpers/config_types.dart';
+import 'package:source_gen/source_gen.dart';
+
+/// Returns a [Builder] for use within a `package:build_runner`
+/// `BuildAction`.
+///
+/// The default Dart code formatter is used.
+Builder scaleCodecPartBuilder({ScaleCodecSerializable? config}) {
```
pastre, line 11: Missing tests

pastre (changes requested):

packages/scale_codec_annotation/lib/src/scale_codec_key.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:meta/meta_meta.dart';
+
+/// An annotation used to specify how a field is serialized
+@Target({TargetKind.field, TargetKind.getter})
+class ScaleCodecKey {
+  /// `true` if the generator should ignore this field completely.
+  ///
+  /// If `null` (the default) or `false`, the field will be considered for
+  /// serialization.
+  final bool? ignore;
```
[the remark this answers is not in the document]
gabrielokura: About the naming topic: very interesting! I found this [ref](https://dart.dev/guides/language/effective-dart/design) very helpful.

packages/scale_codec_serializable/analysis_options.yaml
```diff
@@ -0,0 +1,30 @@
...
+# If you want a smaller set of lints you can change this to specify
+# 'package:lints/core.yaml'. These are just the most critical lints
+# (the recommended set includes the core lints).
+# The core lints are also what is used by pub.dev for scoring packages.
+
+include: package:lints/recommended.yaml
+
+# Uncomment the following section to specify additional rules.
+
+# linter:
```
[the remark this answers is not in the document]
gabrielokura: I don't think so. Dart analyzer is powerful and enough for now.

packages/scale_codec_serializable/lib/src/scale_codec_serializable_generator.dart
gabrielokura: i didn't understand this point

packages/scale_codec_serializable/lib/src/scale_codec_serializable_generator.dart
gabrielokura: Nope, will throw and the build_runner will stop showing the custom error message

packages/scale_codec_annotation/lib/src/scale_codec_key.dart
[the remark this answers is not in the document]
pastre: Yep! According to the bool section, this should be `isIgnored`

packages/scale_codec_serializable/lib/src/scale_codec_serializable_generator.dart
pastre: The `!` operator always throws when `element.library` is `null`. That's unexpected behavior (a function suddenly throwing an error) and exactly the problem that [null safety](https://dart.dev/null-safety/understanding-null-safety) aims to mitigate. You should probably safely unwrap `element.library` before accessing it's value

packages/scale_codec_serializable/test/class_generation_test.dart
```diff
@@ -0,0 +1,31 @@
...
+    'ValidClassWithoutMethods',
+    'ValidClass',
+  };
+
+  final reader = await initializeLibraryReaderForDirectory(
+    p.join('test', 'inputs'),
+    'class_generation_test_input.dart',
+  );
+
+  testAnnotatedElements(
```
pastre, line 26: Is this enough to ensure positive and negative behavior? (Code behaves as we expect and fails as we expect?)

packages/scale_codec_serializable/test/encoded_fields_test.dart
```diff
@@ -0,0 +1,32 @@
...
+    'NumberField',
+    'SetField',
+  };
+
+  final reader = await initializeLibraryReaderForDirectory(
+    p.join('test', 'inputs'),
+    'encoded_fields_test_input.dart',
+  );
+
+  testAnnotatedElements(
```
pastre, line 27: Same as above

packages/scale_codec_serializable/test/inputs/class_generation_test_input.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:scale_codec_annotation/scale_codec_annotation.dart';
```
pastre, line 1: Tests should be readable, clear and concise. Maybe notate which scenario each test is covering and using a GWT pattern to help document the test

packages/scale_codec_serializable/test/inputs/class_generation_test_input.dart
```diff
@@ -0,0 +1,86 @@
...
+}
+
+@ShouldGenerate('''extension _\$ValidClassExtension on ValidClass {
+  String encode() => '';
+
+  ValidClass decode(String encodedData) => ValidClass();
+}
+''')
+@ScaleCodecSerializable(
+    shouldCreateEncodeMethod: true, shouldCreateDecodeMethod: true)
```
pastre, line 83: Aren't we missing two other scenarios? `true` + `false` and `false` + `true`?

packages/scale_codec_serializable/test/inputs/encoded_fields_test_input.dart
```diff
@@ -0,0 +1,144 @@
...
+}
+
+@ShouldGenerate('''extension _\$SetFieldExtension on SetField {
+  String get exampleEncoded => 'TODO: encode example';
+}
+''')
+@ScaleCodecSerializable(
+  shouldCreateDecodeMethod: false,
+  shouldCreateEncodeMethod: false,
+)
```
pastre, line 139: Also missing scenarios

pastre (changes requested):

pastre:
Also please consider splitting this into multiple PRs

packages/scale_codec_annotation/lib/src/scale_codec_serializable.dart
```diff
@@ -0,0 +1,25 @@
...
+
+  /// Creates a new [ScaleCodecSerializable] instance.
+  ///
+  /// Used to load data from [BuilderOptions] on `build.yaml`.
+  factory ScaleCodecSerializable.fromMap(Map<String, dynamic> map) {
+    return ScaleCodecSerializable(
+      shouldCreateDecodeMethod: map['should_create_decode_method'] as bool?,
+      shouldCreateEncodeMethod: map['should_create_encode_method'] as bool?,
+    );
+  }
```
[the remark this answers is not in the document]
gabrielokura: This is veeery interesting. Thank you for the suggestion.

packages/scale_codec_serializable/example/example.dart
```diff
@@ -0,0 +1,13 @@
...
+part 'example.g.dart';
+
+@ScaleCodecSerializable()
+class Person {
+  final String firstName;
+  final String lastName;
+  final bool isYoung;
+
+  Person(
+      {required this.firstName, required this.lastName, this.isYoung = false});
```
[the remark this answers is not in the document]
gabrielokura: The `required` keyword is a way to use the same declaration style of optional parameters (firstName: "example") and still make them mandatory.

For example: 
``` Person(firstName: "example"). ```
You can only declare `firstName: "example"` when you make the parameters optionals. But if you want to ensure they will be passed you should use `required`. 
The compiler will show an error if you don't pass. 

It's nice when you have a class with N parameters, some mandatory, some optionals or nullable and want to keep the `var: value` declaration.

File: packages/scale_codec_serializable/example/example.dart [hunk 19]
```diff
@@ -0,0 +1,13 @@
+import 'package:scale_codec_annotation/scale_codec_annotation.dart';
+
+part 'example.g.dart';
```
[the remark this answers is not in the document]
gabrielokura: Yeah, we will provide an example. I'll add this in the next PR's

packages/scale_codec_serializable/lib/src/config.dart
```diff
@@ -0,0 +1,43 @@
...
+    required this.shouldIgnore,
+  });
+}
+
+/// Represents values from [ScaleCodecSerializable] when merged with local
+/// configuration.
+///
+/// Values are all known, so types are non-nullable.
+class ClassConfig {
+  final bool shouldCreateDecodeMethod;
```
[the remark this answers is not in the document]
gabrielokura: Nit and imo are abbreviation of what words? :)

packages/scale_codec_serializable/lib/src/templates/all_params_template.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'param_template.dart';
+
+class AllParamsTemplate {
+  final Iterable<Param> params;
+
+  const AllParamsTemplate(this.params);
+
+  String generateParams() => _generateParams((param) => param.generateParam());
+
+  String generateParamsUsage() {
```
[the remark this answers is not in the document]
gabrielokura: `generateParamsUsage` is used in `DecodeMethodTemplate` and will be tested in PR #26

packages/scale_codec_serializable/lib/src/templates/class_template.dart
```diff
@@ -0,0 +1,28 @@
+import '../models/scale_codec_class.dart';
+import 'all_params_template.dart';
+import 'decode_method_template.dart';
+import 'encode_method_template.dart';
+
+class ClassTemplate {
```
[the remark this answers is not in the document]
gabrielokura: Class template are tested in `tests/inputs/class_generation_test_input.dart`

packages/scale_codec_serializable/lib/src/templates/decode_method_template.dart
```diff
@@ -0,0 +1,21 @@
+import '../models/scale_codec_class.dart';
+import 'all_params_template.dart';
+
+class DecodeMethodTemplate {
```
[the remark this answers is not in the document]
gabrielokura: Changed to extension. 

And the tests will be created in PR of #26

packages/scale_codec_serializable/lib/src/templates/encode_method_template.dart
```diff
@@ -0,0 +1,16 @@
...
+  final ScaleCodecClass scaleCodecClass;
+
+  const EncodeMethodTemplate(this.scaleCodecClass);
+
+  String generate() {
+    if (!scaleCodecClass.shouldCreateEncodeMethod) {
+      return '';
+    }
+    return '''
+    String encode() => '';
```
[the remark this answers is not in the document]
gabrielokura: Yess, this method will be implemented in PR of #25.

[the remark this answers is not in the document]
pastre: Cool! Might be a good idea to mark those as TODOs, so that we can use something like [TodoBot](https://github.com/apps/todobot) to open issues on remaining TODOs and keep track of those

packages/scale_codec_serializable/lib/src/config.dart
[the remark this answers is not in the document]
pastre: nit = nitpicking (picuinha, super atenção a detalhe)
imo = in my opinion

packages/scale_codec_serializable/example/example.dart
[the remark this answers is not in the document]
pastre: Cool, thanks! :)

packages/scale_codec_serializable/lib/src/templates/all_params_template.dart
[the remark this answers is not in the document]
pastre: We should always ship our code with tests

packages/scale_codec_serializable/test/class_generation_test.dart
gabrielokura: For now, our API only throw errors when the `@ScaleCodecSerializable` decorator is used in the wrong way. More possibilities will come with decode and encode methods implementation. 

But i'll add more examples and tests about the incorrect use of the decorator

gabrielokura:
> Also please consider splitting this into multiple PRs

Sure, next PR's will be smaller.

packages/scale_codec_serializable/lib/src/utils/checks.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  final constructor = constructorElements.first;
+
+  if (constructor.isFactory) {
+    throw InvalidGenerationSourceError(
+      '[ERROR] Class ${element.name} cant have one factory constructor',
+      element: element,
+    );
+  }
+}
```
pastre, line 37: These look like they belong in an extension
```suggestion

extension ClassElementCommonCheck on ClassElement {
	void isNotAbstract() {
	  if (isAbstract) {
	    throw InvalidGenerationSourceError(
	      '[ERROR] Class ${name} cannot be abstract',
	      element: this,
	    );
	  }
	}
	
	/// Check if the class has only `one` default `const` constructor.
	///
	/// Used to avoid errors when decoding `Substrate` classes data.
	void hasOneConstConstructor() {
	  final constructorElements = constructors;
	
	  if (constructorElements.length != 1) {
	    throw InvalidGenerationSourceError(
	      '[ERROR] To many constructors in ${name} class. Class can have only one constructor',
	      element: this,
	    );
	  }
	
	  final constructor = constructorElements.first;
	
	  if (constructor.isFactory) {
	    throw InvalidGenerationSourceError(
	      '[ERROR] Class ${name} cant have one factory constructor',
	      element: this,
	    );
	  }
	}
}
```

packages/scale_codec_serializable/test/class_generation_test.dart
pastre: Good APIs don't have a wrong way to be used. That looks like an issue for the future

PR #30, opened 2022-09-30
Change ss58-codec tests

gabrielokura:
# What was done
1. Change encode exception tests to check error message.
2. Change decode exception tests to check error message. 
3. Create new decode test to reach 100% test coverage.


packages/ss58_codec/coverage/coverage.json
```diff
@@ -1 +1 @@
-{"type":"CodeCoverage","coverage":[{"source":"package:ss58_codec/src/exceptions.dart","script":{"type":"@Script","fixedId":true,"id":"libraries/1/scripts/package%3Ass58_codec%2Fsrc%2Fexceptions.dart","uri":"package:ss58_codec/src/exceptions.dart","_kind":"library"},"hits":[18,0,20,0,29,1,34,1,36,4,4,1,9,1,11,4]},{"source":"package:ss58_codec/src/address.dart","script":{"type":"@Script","fixedId":true,"id":"libraries/1/scripts/package%3Ass58_codec%2Fsrc%2Faddress.dart","uri":"package:ss58_codec/src/address.dart","_kind":"library"},"hits":[15,1,17,1,18,3,20,1,22,3]},{"source":"package:ss58_codec/src/ss58_codec_base.dart","script":{"type":"@Script","fixedId":true,"id":"libraries/1/scripts/package%3Ass58_codec%2Fsrc%2Fss58_codec_base.dart","uri":"package:ss58_codec/src/ss58_codec_base.dart","_kind":"library"},"hits":[35,1,36,2,37,2,38,1,42,2,43,1,45,2,51,5,52,4,53,5,56,1,60,2,61,1,62,1,65,1,66,1,67,1,68,1,72,1,75,1,76,2,77,6,82,1,84,3,78,0,96,1,97,1,98,2,99,1,101,2,104,1,105,1,106,1,107,1,110,1,111,1,115,2,119,1,120,3,121,1,127,3,130,1,131,7,135,7,139,2,140,1,141,2,142,4,144,2,12,3,17,2,151,1,152,1,155,1,158,2,160,5,163,1,165,2,18,1]}]}
```
justkawal: please put coverage folder in `.gitignore` inside `ss58_codec`
Like: `coverage/`

justkawal (changes requested):

packages/ss58_codec/test/ss58_codec_test.dart
```diff
@@ -56,78 +56,178 @@ void main() {
     }
   });
 
-  // Decode Exception testing
+  /// [SS58Codec] `decode` method Exception testing
   group('decode: throw exception -> ', () {
+    /// `String` address with `length < 3`
     test('data.length < 3 ', () {
```
pastre, line 62: I strongly recommend[ BDD Unit test naming conventions ](https://matheus.ro/2017/09/24/unit-test-naming-convention/) when naming tests. Tests should be readable and have a well defined behavior. This applies to all unit tests in this PR lol
```suggestion
    test('Should throw exception when data lenght is too small', () {
```

packages/ss58_codec/test/ss58_codec_test.dart
```diff
@@ -56,78 +56,178 @@ void main() {
...
+        ),
+      );
     });
   });
 
-  // Encode Exception testing
+  /// [SS58Codec] `encode` method Exception testing
   group('encode: throw exception -> ', () {
+    /// [Address] with invalid `prefix`
     test('invalid prefix ', () {
```
pastre, line 147: Using BDD's naming convention will also clean up some of your comments. They are not bad, but are usually a code smell. Your code should read itself, and if you're having to explain it using comments, usually it's not readable.
Also, use `///` only for documentation. You can use common `//` for comments
```suggestion
    test('it should throw InvalidPrefixException when an address has invalid prefix', () {
```

packages/ss58_codec/test/ss58_codec_test.dart
```diff
@@ -56,78 +56,178 @@ void main() {
...
-          throwsA(isA<InvalidPrefixException>()));
+        () => SS58Codec.encode(addressWithNegativePrefix),
+        throwsA(
+          predicate((exception) =>
+              exception is InvalidPrefixException &&
+              exception.toString() == 'Invalid SS58 prefix byte: -1.'),
+        ),
+      );
+
+      /// [Address] invalid because of prefix is greater than the limit
```
pastre, line 160: You should break this into multiple tests. Good unit tests usually fail for a single reason (that's very utopic, but nice to have). You can have ` test('it should throw InvalidPrefixException when an address has negative prefix'` and ` test('it should throw InvalidPrefixException when prefix is greater than the limit'`

packages/ss58_codec/test/ss58_codec_test.dart
```diff
@@ -56,78 +56,178 @@ void main() {
...
+          () => SS58Codec.encode(invalidAddress),
+          throwsA(
+            predicate(
+              (exception) =>
+                  exception is BadAddressLengthException &&
+                  exception.toString() ==
+                      'Bad Length Address: ${invalidAddress.toString()}.',
+            ),
+          ),
+        );
```
pastre, line 188: nit: tests intrinsically have an expected value and an actual value. Notating those help with the test's readability. I'd also hard code `${invalidAddress.toString()}` to avoid changes on string conversion to affect the outcome of this test
```suggestion
         
          final expectedErrorMessage = 'Bad Length Address: ${invalidAddress.toString()}.' // TODO hardcode  ${invalidAddress.toString()}
          () => SS58Codec.encode(invalidAddress),
          throwsA(
            predicate(
              (exception) =>
                  exception is BadAddressLengthException &&
                  exception.toString() == expectedErrorMessage ,
            ),
          ),
        );
```

pastre (changes requested):
Small improvements on tests. Overall looking good!
You forgot to remove `coverage.json` and `lcov.info` from your branch xD

packages/ss58_codec/test/ss58_codec_test.dart
justkawal: I see, but as far as I know here its `pass by value` and `not pass by reference` so variable -> `invalidAddress` is never going to change itself or from within the encode function call.

pastre: I agree that it's not going to change functionality, but we're writing tests. Readability and maintainability are really really super important to ensure that we don't have any bug in our tests (bc they are the source of truth deciding whether or not our code works. Finding bugs there is terrible) and they are not a pain to maintain. To the compiler it's the same, but it's a huge difference for us. As far as the compiler knows, this one-liner is good enough lol `() => SS58Codec.encode(invalidAddress),throwsA(predicate((exception) => exception is BadAddressLengthException && exception.toString() == 'Bad Length Address: ${invalidAddress.toString()}.' ,),),);`

justkawal: Alright haha, will try to maintain as much readability as possible.

packages/ss58_codec/test/ss58_codec_test.dart
leonardocustodio: Highly agree with this. We should be able to know what is the goal of the tests only by their name. Even if the name is a bit big.

leonardocustodio (changes requested):

packages/ss58_codec/test/ss58_codec_decode_test.dart
```diff
@@ -0,0 +1,141 @@
...
+void main() {
+  group('SS58Codec decode method tests', () {
+    test('Should decoded data be equal to expected address', () {
+      final Address decodedAddress =
+          SS58Codec.decode('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY');
+
+      final expectedAddress = Address(
+          prefix: 42,
+          bytes: Uint8List.fromList(hex.decode(
+              'd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d')));
```
pastre, line 19: nit: avoid magic values. Consider creating named variables for `42 ` and `Uint8List.fromList(hex.decode(
              'd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'))`. When creating variables in a test, they usually have a well defined role. Most of the times, the role is either one the five described in [this link](https://martinfowler.com/bliki/TestDouble.html). Testing doubles is the next step to upgrade your tests to the next level :)

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+    test('Should encode all addresses sucessfully', () {
+      final address1 =
+          Address(prefix: 0, bytes: Uint8List.fromList([1, 2, 3, 4]));
+      final address2 = Address(prefix: 64, bytes: Uint8List.fromList([1, 2]));
+      final address3 = Address(prefix: 16383, bytes: Uint8List.fromList([2]));
+
+      expect(() => SS58Codec.encode(address1), returnsNormally);
+      expect(() => SS58Codec.encode(address2), returnsNormally);
+      expect(() => SS58Codec.encode(address3), returnsNormally);
+    });
```
pastre, line 20: Consider creating multiple tests for each scenario
`test('Should encode when prefix is zero',`
`test('Should encode when prefix is 64',`
`test('Should encode when prefix is 16383',`.

Also, why those values?

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        () {
+      final originalAddress = Address(
+          prefix: 64, bytes: Uint8List.fromList(List<int>.filled(32, 32)));
+
+      final String encodedAddress = SS58Codec.encode(originalAddress);
+
+      final Address decodedAddress = SS58Codec.decode(encodedAddress);
+
+      decodedAddress.isEqual(originalAddress);
+    });
```
pastre, line 33: You're depending on `SS58Codec.encode`. Good unit tests have only a single reason to fail, and a flaky bug on `encode` will result on this test being flaky. Please consider hardcoding the setup:
Also: isn;t this missing an assertion?
```suggestion
    test(
        'Should encode the address and then re-decode that encoded object to match with original bytes passed',
        () {
      final String encodedAddress = SS58Codec.encode("originalAddress"); // TODO get encoded original address
      final Address decodedAddress = SS58Codec.decode(encodedAddress);
      decodedAddress.isEqual(originalAddress);
    });
```

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+              exception is InvalidPrefixException &&
+              exception.toString() == expectedErrorMessage),
+        ),
+      );
+    });
+
+    test(
+        'Should throw BadAddressLengthException when an address bytes length is equal to 3, 5, 6 or 7.',
+        () {
+      for (var len in [3, 5, 6, 7]) {
```
pastre, line 72: Please avoid loops in tests. you can create a single test case for each of those
Also, please consider renaming variables for clarity
```suggestion
      for (var valueStub in [3, 5, 6, 7]) {
```

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+            ),
+          ),
+        );
+      }
+    });
+
+    test(
+        'Should throw BadAddressLengthException when an address bytes length is equal to [9-31]',
+        () {
+      for (var len = 9; len <= 31; len++) {
```
pastre, line 93: You don't need to test all the values. Imagine if the range was wider: from 9 to 310000000. Tests would take forever to run. 
You can test : 

- 8 (lower bound - 1, should not throw)
- 9 (lower bound, should throw)
- 31 (upper bound, should throw)
- 32 (upper bound + 1, should throw) 

and that should be enough. Also, please create a different test for each of those

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+            ),
+          ),
+        );
+      }
+    });
+
+    test(
+        'Should throw BadAddressLengthException when an address bytes length is equal or greater than 34',
+        () {
+      for (var len in [34, 35]) {
```
pastre, line 115: Please separate into multiple tests:
- 33 is ok
- 34 breaks
- 35 breaks

packages/ss58_codec/test/ss58_codec_encode_test.dart
```diff
@@ -0,0 +1,138 @@
...
+                  exception.toString() == expectedErrorMessage,
+            ),
+          ),
+        );
+      }
+    });
+  });
+}
+
+int randomInt(int max) {
```
pastre, line 136: This is not being used and can be deleted

packages/ss58_codec/test/ss58_codec_encode_test.dart
gabrielokura: They are common values used just to test the method

gabrielokura: We have two situations of success (not including exceptions cases) in the encode function: 
- for prefix < 64
- and for prefix > 64

So, i remove the `test('Should encode when prefix is 16383') and create tests for bytes.length that were missing.

packages/ss58_codec/test/ss58_codec_encode_test.dart
gabrielokura: Thanks. I see that i was taking the wrong approach for these tests. 

I will test only the edge cases of `SS58Codec.encode()`

packages/ss58_codec/test/ss58_codec_decode_test.dart
```diff
@@ -0,0 +1,141 @@
...
+
+      // Match whether the 2 addresses are equal or not
+      decodedAddress.isEqual(expectedAddress);
```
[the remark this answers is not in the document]
gabrielokura: It is. We've used extensions on `Address` class to make the assertion. 

But i agree this approach is not so readable.

pastre (approved):

leonardocustodio (approved):

PR #43, opened 2022-10-04
Update ss58 test cases

gabrielokura:
# What was done

1. Update ss58 base tests .
2. Create registry item tests file. 
3. Create registry tests file.

# Next PR

Will be incrementing ss58 documentation and example folder.


packages/ss58/example/ss58_example.dart
```diff
@@ -1,21 +1,23 @@
...
-    print('decoded bytes: $decodedBytes');
-    print('original address: $encodedAddress');
-    print('encoded address:  $encodedAddress');
-    print('decoding and encoding back and forth is awesome');
-  } else {
-    throw Exception('Something went wrong');
-  }
+  final int substrateAddressPrefix = 42;
+  final encodedAddress = Codec(substrateAddressPrefix).encode(decodedBytes);
+  print(encodedAddress);
```
pastre, line 22: Please remove `print` statements

packages/ss58/lib/src/ss58_base.dart
```diff
@@ -4,19 +4,32 @@ import 'package:ss58/src/registry.dart';
...
-  /// Initialize Codec from the network prefix
+  /// Initialize Codec from the address prefix.
+  ///
+  /// ```
+  /// [Exceptions]
+  /// throw InvalidPrefixException: when (prefix < 0 || prefix > 16383)
+  /// ```
   Codec(this.prefix) {
-    if (prefix <= 0 || prefix > 16383) {
+    if (prefix < 0 || prefix > 16383) {
```
pastre, line 22: shouldn't we update tests for this?

packages/ss58/lib/src/ss58_base.dart
```diff
@@ -4,19 +4,32 @@ import 'package:ss58/src/registry.dart';
...
       throw InvalidPrefixException(prefix);
     }
   }
 
-  /// Initialize Codec from network name
+  /// Returns a [Registry] instance from `ss58-registry-json`
+  /// with all known SS58 account types.
+  static final registry = Registry.fromJsonString(reg.jsonRegistryData);
```
pastre, line 29: nit: better naming indicating that it returns a registry with all known SS58 account types.

packages/ss58/example/ss58_example.dart
gabrielokura: Isn't it good for developers observe when they run this example?

packages/ss58/lib/src/ss58_base.dart
gabrielokura: Added test case for prefix = 0

packages/ss58/lib/util/ss58_registry_json.dart
```diff
@@ -1,5 +1,8 @@
 ///
-/// Data of para-chains
+/// Data of para-chains.
+///
+/// Updated in 04/08/2022 from https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json
+// TODO: load ss58-registry.json from paritytech github
```
Lohann, line 5: Nice! could you create a ticket for that? so we can remind us in the future to automatize this somehow.

leonardocustodio (approved):

PR #45, opened 2022-10-05
Add CodeClimate

leonardocustodio:


pastre:
@leonardocustodio  Aren't we using CodeCov? Never seen this tool, is it different?

leonardocustodio:
Yeah, I'm going to replace codecov with codeclimate....
Codecov is only focused on coverage, while climate has other quality tools embedded on it.
Though I've not finished the configuration yet, I will finish it at night.

PR #53, opened 2022-10-05
Adds pull request template

pastre:
## Any follow ups?

Insert a reference to a pull request that should be reviewed before this is reviewed

## Description
Adds a pull request template that should be loaded when creating a PR

## Why?

1. Create a pattern on what's needed to open a PR
2. It's easier to fill a template then to think what to write every time
3. Ease reviews

## How?

Created a file with a template

## Testing Plan 
Not applicable


justkawal (approved):
LGTM

Issue #40, opened 2022-10-03
Re-write Unit Tests for BigInt


justkawal:
For functions file re-write BigInt Unit Test-Cases:
- [x] Remove for loops.
- [x] Remove `Random()` values usage.
- [x] Proper Description of test cases with readability being the most important.
- [x] Separately test Happy and Bad cases in different groups.

PR #62, opened 2022-10-05
Fix/rewrite test cases bigint

justkawal:
## Any follow ups?
Please review PR: #22 before reviewing this PR.

## Description
This PR fixes: #40 by adding test cases to test BigInt functionality inside functions file.
- Tested Happy Cases.
- Tested Edge and Corner Cases.
- Tested Cases where Exception is to be thrown.
- Tested Extension methods which converts **int** and **String** to **BigInt**.


gabrielokura (dismissed):
nit: you don't need to use equals() with expect function. You can just write:
``` 
expect(toSignedBigInt(lowestValue, 64).toInt(), lowestValue);
```

gabrielokura:
Also, just passing along what @pastre said to me early: we don't need to test the type of the exception and it's message every time. 

We could just test the exception message in a single unit test and keep the others with the type check.

justkawal:
> Also, just passing along what @pastre said to me early: we don't need to test the type of the exception and it's message every time.
> 
> We could just test the exception message in a single unit test and keep the others with the type check.

@leonardocustodio said to not only test the happy cases where the code passes and also to consider testing the cases where the code breaks.

He is right about it as @Lohann also suggest to test the cases which can break your code and be sure that it's breaking and not passing.

gabrielokura:
> > Also, just passing along what @pastre said to me early: we don't need to test the type of the exception and it's message every time.
> > We could just test the exception message in a single unit test and keep the others with the type check.
> 
> @leonardocustodio said to not only test the happy cases where the code passes and also to consider testing the cases where the code breaks.
> 
> He is right about it as @Lohann also suggest to test the cases which can break your code and be sure that it's breaking and not passing.

I'm not talking about testing exception or not. We `must` create unit tests for exceptions. 

I'm just saying that you can create a single unit test for the message: 
```
expect(..., throwsA(predicate(exception) => exception is ExceptionType && exceptio.toString() == ...));
``` 

And the other can be: 
```
expect (..., throwsA(isA<ExceptionType>()));
```

justkawal:
> > > Also, just passing along what @pastre said to me early: we don't need to test the type of the exception and it's message every time.
> > > We could just test the exception message in a single unit test and keep the others with the type check.
> > 
> > 
> > @leonardocustodio said to not only test the happy cases where the code passes and also to consider testing the cases where the code breaks.
> > He is right about it as @Lohann also suggest to test the cases which can break your code and be sure that it's breaking and not passing.
> 
> I'm not talking about testing exception or not. We `must` create unit tests for exceptions.
> 
> I'm just saying that you can create a single unit test for the message:
> 
> ```
> expect(..., throwsA(predicate(exception) => exception is ExceptionType && exceptio.toString() == ...));
> ```
> 
> And the other can be:
> 
> ```
> expect (..., throwsA(isA<ExceptionType>()));
> ```

As per my opinion and I feel it's a better approach, I think it should be better if we do not write another expect again.

Lohann (approved):

leonardocustodio (dismissed):

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:test/test.dart';
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+
+void main() {
+  group('IntConverter Extension:', () {
+    test(
+        '5 as int will be converted to BigInt.from(5) when used with IntConverter Extension',
```
pastre, line 7: Please consider using [BDD Naming conventio](https://matheus.ro/2017/09/24/unit-test-naming-convention/)n when creating your tests: **Given_Preconditions_When_StateUnderTest_Then_ExpectedBehavior **
```suggestion
        'Given any positive integer, it should convert to BigInt ',
```

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:test/test.dart';
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+
+void main() {
+  group('IntConverter Extension:', () {
```
pastre, line 5: Needed cases for this are:

- Positive integers
- Zero
- Negative Integers
- Integers larger then BigInt support (not sure if there's one in Dart)
- Integers smaller then BigInt support (not sure if there's one in Dart)

I'm only seeing two positive ints

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
```diff
@@ -0,0 +1,42 @@
...
+    test(
+        '5 as int will be converted to BigInt.from(5) when used with IntConverter Extension',
+        () {
+      final bigInt = 5.toBigInt;
+      expect(BigInt.from(5), equals(bigInt));
```
pastre, line 10: nit: don't use the library to test itself: consider parsing your actual value into another data type (like String) in order to assert

justkawal: It's not library, its internal dart's pure function from dart sdk.

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
justkawal: Thanks, for pointing it out, I'll add test-cases for negative, zero, positive integers.

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
justkawal: I'll match the ending type as Strings in `expect` clause.

pastre: The reason for this is that we don't want changes in BigInt impacting the expectation of our test: only the behavior

justkawal: Okay brother.

pastre (changes requested):
Comments apply to multiple points throughout  your PR :)

pastre:
> > > > Also, just passing along what @pastre said to me early: we don't need to test the type of the exception and it's message every time.
> > > > We could just test the exception message in a single unit test and keep the others with the type check.
> > > 
> > > 
> > > @leonardocustodio said to not only test the happy cases where the code passes and also to consider testing the cases where the code breaks.
> > > He is right about it as @Lohann also suggest to test the cases which can break your code and be sure that it's breaking and not passing.
> > 
> > 
> > I'm not talking about testing exception or not. We `must` create unit tests for exceptions.
> > I'm just saying that you can create a single unit test for the message:
> > ```
> > expect(..., throwsA(predicate(exception) => exception is ExceptionType && exceptio.toString() == ...));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And the other can be:
> > ```
> > expect (..., throwsA(isA<ExceptionType>()));
> > ```
> 
> As per my opinion and I feel it's a better approach that if we do not write another expect again, as it's the same.

We should have a single assertion testing a single condition. It's easier to debug and find the cause of a failing test, and also makes it clearer to understand how the test is asserting its requirements. We should **NOT** worry about code duplication in our tests: focus on readability and maintainability

packages/polkadart_scale_codec/test/functions_test/to_signed_bigint_test.dart
```diff
@@ -0,0 +1,193 @@
...
+
+  // This test will throw `UnexpectedCaseException` when unknow bit-size is passed as an argument
+  {
+    group('UnexpectedCaseException', () {
+      final exceptionMessage = 'Unexpected BitSize: 20.';
+      test(
+          'toSignedBigInt will throw \'UnexpectedCaseException\' when unknow bit-size: \'20\' is passed as an argument',
+          () {
+        expect(
+            () => toSignedBigInt(200, 20),
```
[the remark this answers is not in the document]
justkawal: But we had to pass it as a sort of function to make it throw and catch.
It's the way how we match exceptions.

You can find it in: https://pub.dev/packages/test

.
Screenshot 2022-10-08 at 12 48 47 AM

packages/polkadart_scale_codec/test/functions_test/to_signed_bigint_test.dart
```diff
@@ -0,0 +1,193 @@
...
+            throwsA(
+              predicate((e) =>
+                  e is InvalidSizeException &&
+                  e.toString() == exceptionMessage),
+            ));
+      });
+    });
+  }
+
+  // This test will throw `UnexpectedTypeException`
```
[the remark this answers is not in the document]
justkawal: Just used for more clarity.
I found it to be useful for to spot quickly the new starting sections easily.

packages/polkadart_scale_codec/test/functions_test/to_signed_bigint_test.dart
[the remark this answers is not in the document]
pastre: I understand, my suggestion is to make your test clearer, not to change assertion matching. If you can't avoid the one-liner, it's fine. But it's usually a code smell in your tests
Please pay attention to the naming: `stub` and `dummy` are not random names, they follow a well-known test pattern called [Testing Doubles](https://martinfowler.com/bliki/TestDouble.html)
```suggestion
			// Given
			final decodedNumberDummy = 200;
			final bitBaseStub = 20;
			
			// When / Then
			 expect( () => decodedNumberDummy(decodedNumberDummy, bitBaseStub), throwsA(isA<UnexpectedCaseException>()));
```

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
pastre: Interesting, why'd `dprint` block the pipeline?

packages/polkadart_scale_codec/test/functions_test/to_signed_bigint_test.dart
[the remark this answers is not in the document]
justkawal: Alright got it.

justkawal:
@pastre Requested changes resolved.

gabrielokura (dismissed):
LGTM

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
```diff
@@ -3,40 +3,53 @@ import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
...
-    test(
-        '500 as int will be converted to BigInt.from(500) when used with IntConverter Extension',
+
+    test('When provided with 0 as int, it will result in an BigInt Object.',
+        () {
+      final bigInt = 0.toBigInt;
+      expect(BigInt.from(0), equals(bigInt));
+    });
+
+    test('When provided with 500 as int, it will result in an BigInt Object.',
```
pastre, line 18: Isn't this testing the same as `When provided with 5 as int, it will result in an BigInt Object.`?
Also, our descriptions should be about the code's behavior, not it's structure. `BigInt` is a tool we're using for handling with large objects: it's related to structure. `500` is a positive integer we're using to validate a test case, it's releated to the structure .

**This comment applies to all test messages** (I can point each of them out if that helps)
```suggestion
    test('When provided with a positive integer, it should convert to a big integer',
```

packages/polkadart_scale_codec/test/functions_test/extensions_test.dart
pastre: This wasn't solved, but it's a nit

packages/polkadart_scale_codec/test/functions_test/to_signed_bigint_test.dart
```diff
@@ -68,125 +102,92 @@ void main() {
...
+        final value = BigInt.from(429496726);
+
+        expect(() => toSignedBigInt(value, bitsize),
+            throwsA(isA<UnexpectedTypeException>()));
       });
     });
   }
 
-  // This test will throw `UnexpectedCaseException` when unknow bit-size is passed as an argument
+  // When unknow bit-size is passed as an argument it should throw UnexpectedCaseException
```
pastre, line 180: You can remove the comment. The test is self-explanatory now and needs no documentation - which is very good :)

pastre (commented):
Couple comments, none of them blocking :)

gabrielokura (approved):

pastre (approved):

Issue #54, opened 2022-10-05
Test Cases for Signed and Unsigned Int


justkawal:
Tasks:
- [ ] Happy Cases
- [ ] Edge Cases
- [ ] Cases which throw exceptions
- [ ] Testing with different value types
- [ ] Testing with (lowest - 1) and (highest + 1) values

PR #73, opened 2022-10-07
Test cases for encode Signed and Unsigned Integers

gabrielokura:
## Description
Creating unit tests for encoding signed and unsigned integers. 
Creating tests for `functions.dart` to achieve greater tests coverage. 

## Why?
Issue #54 

## How?
1. Separate unsigned integers test file.
2. Create unsigned integers unit tests.
3. Separate signed integers test file.
4. Create signed integers unit tests.
5. Create util/functions_test.dart
6. Fixing some methods in `functions.dart` and create unit tests.

## Testing Plan 
Run `dart test` on packages/polkadart_scale_codec folder.


packages/polkadart_scale_codec/lib/src/util/functions.dart
```diff
@@ -157,13 +157,13 @@ bool checkUnsignedInt(int val, int bitSize) {
...
     case 16:
-      max = 0xffff;
+      max = 0x7fff;
       break;
     case 32:
-      max = 0xffffffff;
+      max = 0x7fffffff;
       break;
     default:
       throw UnexpectedCaseException('Unexpected BitSize: $bitSize.');
```
pastre, line 169: Do we have a test for this exception?

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
```diff
@@ -0,0 +1,182 @@
+import 'package:polkadart_scale_codec/src/core/core.dart';
+import 'package:polkadart_scale_codec/src/util/utils.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('encode to hex signed int tests', () {
```
pastre, line 6: Great idea to use groups to improve organization!

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
```diff
@@ -0,0 +1,182 @@
...
+    registry.select('Codec');
+
+    // fetching the parsed types from `Json` to `Type`
+    final types = registry.getTypes();
+
+    // Initializing Scale-Codec object
+    final codec = Codec(types);
+
+    group('fixed-length 8bit', () {
+      test('Should encode and return correct value ', () {
```
pastre, line 27: Please improve titles. Same for other tests
```suggestion
      test('Given a positive integer when it is within range it should be encoded', () {
```

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
```diff
@@ -0,0 +1,182 @@
...
+    // specifyng which schema type to use
+    registry.select('Codec');
+
+    // fetching the parsed types from `Json` to `Type`
+    final types = registry.getTypes();
+
+    // Initializing Scale-Codec object
+    final codec = Codec(types);
+
+    group('fixed-length 8bit', () {
```
pastre, line 26: Please add a special case for zero

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
```diff
@@ -0,0 +1,182 @@
...
+          throwsA(isA<InvalidSizeException>()),
+        );
+      });
+
+      test(
+          'Should throw InvalidSizeException when value is greater than 2147483647',
+          () {
+        int value = 2147483648;
+        int registryIndex = registry.getIndex('i32');
+
```
pastre, line 174: nit: use `const` and `final`  when possible. It's a hint for devs that that value should not change
nit: avoid declaring types when possible. Let the compiler help you with that

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,142 @@
+import 'package:polkadart_scale_codec/src/util/utils.dart';
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  group('functions checkSignedInt()', () {
+    test('Should returns normally when bitSize is 8', () {
```
pastre, line 7: Many ways to improve this :)
```suggestion
    test('Given bitSize is 8 and value is within range check should pass', () {
```

```suggestion
    test('Given bitSize is valid and value is within range check should not throw', () {
```

```suggestion
    test('Given bitSize is 8 and value is valid check should not fail', () {
```

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,142 @@
...
+
+    test('Should returns normally when bitSize is 16', () {
+      int bitSize = 16;
+      int value = 10;
+
+      expect(() => checkSignedInt(value, bitSize), returnsNormally);
+    });
+
+    test('Should returns normally when bitSize is 32', () {
+      int bitSize = 16;
```
pastre, line 22: You're testing for 32 on the description and using 16 in your test. I think you forgot to change when you pasted :p

We should **always** see a failing test before implementing it. This is called the [Red-Green-Refactor](https://www.codecademy.com/article/tdd-red-green-refactor) cycle and can be applied outside of TDD

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,142 @@
...
+      int bitSize = 15;
+      int value = 10;
+
+      final expectedErrorMessage = 'Unexpected BitSize: 15.';
+
+      expect(
+        () => checkSignedInt(value, bitSize),
+        throwsA(
+          predicate((exception) =>
+              exception is UnexpectedCaseException &&
```
pastre, line 40: Just `isA<UnexpectedCaseException>` is enough

packages/polkadart_scale_codec/lib/src/util/functions.dart
justkawal: Yes, we have that in #62.

pastre: Thank you! :)

justkawal: Anytime brother.

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
justkawal: @pastre I think, If the range of acceptable values varies from ( -ve to +ve ) integers then, it would obviously accept **zero** anyhow !!

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,279 @@
...
+
+      final int bitSize = 16;
+      final String sign = 'u';
+      final BigInt value = BigInt.from(100);
+
+      expect(checkBigInt(value, sign, bitSize, min, max), true);
+    });
+
+    test('Should throw InvalidSizeException when value is out of range', () {
+      final BigInt min = BigInt.from(0);
```
justkawal, line 260: Good one @gabrielokura, specifying types is whats promoted more by dart analyser and is something which the analyser suggests as a quick-fix. 

I becomes more clear to the user about the type it's holding and don't have to hover over the function to see the info about it.

packages/polkadart_scale_codec/test/test_types/signed_int_test.dart
pastre: Zero is a safe approach to avoid confusion with `null`, as it has been used historically to represent that

justkawal: That was a long-time ago, anyways its ok.

packages/polkadart_scale_codec/test/util/functions_test.dart
gabrielokura: A new test file for all exceptions and messages is necessary or useless? 

If i remove the message checker the test coverage will decrease together

leonardocustodio (approved):

packages/polkadart_scale_codec/test/util/functions_test.dart
gabrielokura: nice, i got it! Thanks

Lohann (approved):

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,275 @@
...
+        throwsA(
+          predicate((exception) =>
+              exception is UnexpectedCaseException &&
+              exception.toString() == expectedErrorMessage),
+        ),
+      );
+    });
+  });
+
+  group('functions checkInt()', () {
```
pastre, line 166: Couple refactors on `checkInt`:
1. We don't need `checkInt` and `checkBigInt`: use generics
```dart
bool compare(Comparable min, Comparable max, Comparable value) {
  return value.compareTo(min) > 0 && value.compareTo(max) < 0;
}
```
2. Its signature should be `checkInt(int val, String sign, int bitSize, int min, int max)`. Don't return `bool` when it's unecessary and don't use 'dynamic' when it's not needed

pastre (commented):
Was reviewing while it was merged

packages/polkadart_scale_codec/test/util/functions_test.dart
```diff
@@ -0,0 +1,275 @@
...
+    test('Should return true when value is in range', () {
+      final int min = 0;
+      final int max = 100;
+
+      final int bitSize = 16;
+      final String sign = 'u';
+      final int value = 50;
+
+      expect(checkInt(value, sign, bitSize, min, max), true);
```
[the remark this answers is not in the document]
gabrielokura: done!

PR #70, opened 2022-10-07
Test cases for Vec

justkawal:
## Any follow ups?
Please review and approve PR: #62, before reviewing this PR.

## Description
* It adds test-cases for `Vec`.
* Combined Test-Cases:
    * `Vec<u8>`
    * `Vec<Vec<u8>>`
    * `Vec<u256>`
    * `Vec<i256>`
    * `Vec<(u32, u32, u16)>`
    * `Vec<(u32, Option<bool>)>`
    * `Vec<(String, bool)>`
    * `Vec<bool>`
    * `Vec<Option<bool>>` 
    * `Vec<Option<u8>>`
    * `Vec<Text>`

## How?
Testing multiple combination of test-cases helps to narrow down the breaking scenarios.

## Testing Plan 
Test the file `vec_test.dart`.

Issue #42, opened 2022-10-03
Sink Unit Test


justkawal:
- [x] Unit Test Sink

Acceptance Criteria:
* Edge Cases
* Happy cases
* Bad Cases where code breaks

PR #74, opened 2022-10-07
Sink Unit Test-Cases

justkawal:
## Any follow ups?
Please review #70, before reviewing this PR.

## Description
* Adds unit-test cases for Sink Class.
* Tests Exceptions.
* Tests lowest and highest acceptable value.
* Abbreviations not removed, as it helps to mirror the function calls.

## Why?
* Issue #42


gabrielokura (commented):
Why do we need `sink_test.dart`?

Looks like we are re-writing primitive types tests under `test/test_types/...`

gabrielokura:
> 

So i think we need to test `write`, `bytes` and `toHex` too, right? @pastre

justkawal:
> So i think we need to test `write`, `bytes` and `toHex` too, right? @pastre

Oh yeah, thanks

leonardocustodio (approved):

gabrielokura (approved):

Lohann (approved):

PR #75, opened 2022-10-08
Starting point of substrate_metadata

justkawal:
This is the starting point of this substrate_metadata repo.

It adds:
* Package and its other necessary files.
* Substrate required types and aliases.


justkawal:
@Lohann, @leonardocustodio needs your review as you guys know alot about substrate and polkadot.

packages/substrate_metadata/lib/old/definitions/substrate/types.dart
```diff
@@ -0,0 +1,3588 @@
+const types = {
```
gabrielokura, line 1: Couldn't this be an interface/class/abstract class?

packages/substrate_metadata/lib/old/definitions/substrate/types_alias.dart
```diff
@@ -0,0 +1,50 @@
+final typesAlias = <String, Map<String, String>>{
```
gabrielokura, line 1: same as above

packages/substrate_metadata/lib/old/definitions/substrate/types.dart
justkawal: Instead I could rename it as SUBSTRATE_TYPES, and it would be much better.

packages/substrate_metadata/lib/old/definitions/substrate/types_alias.dart
justkawal: Instead renaming it as SUBSTRATE_TYPES_ALIAS, and it would be much better.

packages/substrate_metadata/lib/old/definitions/substrate/types.dart
justkawal: keeping it as normal const is better.

leonardocustodio (changes requested):
Add the reference where you got the types and we should be good to go

justkawal:
> Add the reference where you got the types and we should be good to go

Adding references:

`substrateTypes`: https://github.com/subsquid/squid/blob/cc0bc27e16c5c441c39d847be2d560f57afd3b38/substrate-metadata/src/old/definitions/substrate/types.ts#L1

`substrateTypesAlias`: https://github.com/subsquid/squid/blob/cc0bc27e16c5c441c39d847be2d560f57afd3b38/substrate-metadata/src/old/definitions/substrate/typesAlias.ts#L4

leonardocustodio (approved):

Lohann (approved):

packages/substrate_metadata/example/substrate_metadata_example.dart
```diff
@@ -0,0 +1 @@
+void main() {}
```
Lohann, line 1: ??

packages/substrate_metadata/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: substrate_metadata
+description: substrate_metadata description
+version: 1.0.0
+homepage: https://github.com/rankanizer/polkadart
+
+# TODO: Remove this when polkadart_scale_codec is published
```
Lohann, line 6: Please create a issue for that and put here, so we remember to publish it in the future

justkawal: Now trackable with issue: #83 ( substrate_metadata publish configuration )

packages/substrate_metadata/example/substrate_metadata_example.dart
justkawal: Removed unused file ✅.

Issue #41, opened 2022-10-03
Unit Test Source


justkawal:
Unit Test for Compact Functionality:
- [x] Happy Cases
- [x] Edge Cases
- [x] Bad Cases
- [x] Testing with different types

PR #76, opened 2022-10-08
Source Unit Test

justkawal:
## Any follow ups?
Please review #74, before reviewing this PR.

## Description
* Adds unit-test cases for Source Class.
* Tests Exceptions.
* Tests lowest and highest acceptable value.
* Abbreviations not removed, as it helps to mirror the function calls.

## Why?
* Issue #41


leonardocustodio (approved):

packages/polkadart_scale_codec/lib/src/core/source.dart
```diff
@@ -138,7 +138,7 @@ class Source {
   }
 
   int compactLength() {
-    var len = compact();
+    var len = uncompact();
```
Lohann, line 141: why the name of this method is `compactLength` when it doesn't return a length? Also it may return an error when the length is greater than dart's `Int` type.

Lohann (approved):

packages/polkadart_scale_codec/test/source_test.dart
```diff
@@ -0,0 +1,252 @@
...
+        methodName: 'boolean', expectedValue: true, compactedHex: '0x01');
+    // false
+    testPrimitiveUnCompactingFromSource(
+        methodName: 'boolean', expectedValue: false, compactedHex: '0x00');
+
+    testPrimitiveUnCompactingFromSource(
+        methodName: 'str',
+        expectedValue: 'github: justkawal',
+        compactedHex: '0x446769746875623a206a7573746b6177616c');
+  });
```
Lohann, line 251: I'm missing a test for `EOFException` here:

`bytes`: returns an `EOFException` when the bytes length doesn't match
`str`: Same case, and also should return an error if the bytes aren't a valid String

Lohann (changes requested):

packages/polkadart_scale_codec/lib/src/core/source.dart
justkawal: Done ✅.
Renamed to `lengthOfCompact`.

packages/polkadart_scale_codec/test/source_test.dart
justkawal: Added exception test-cases ✅.

packages/polkadart_scale_codec/test/source_test.dart
```diff
@@ -0,0 +1,278 @@
...
+    test(
+        'When invalid Hexa-decimal like: \'0x1700007014057fd8b806\' is passed as an argument then it should throw \'AssertionException\'.',
+        () {
+      final value = '0x1700007014057fd8b806';
+
+      final source = Source(value);
+
+      expect(() => source.str(), throwsA(isA<AssertionException>()));
+    });
+  });
```
Lohann, line 277: **[optional suggestion]**
I don't think this should be an `AssertionException`, instead should be something more granular, like `InvalidStringException`

Lohann (approved):

packages/polkadart_scale_codec/test/source_test.dart
justkawal: Can be a `InvalidCompactLengthException` as it was predicted to get `int` when an encoded string as hex was passed but found `BigInt` in length.

justkawal: Nice I'll change it right-away.

Lohann: What happens if you provide an invalid ut8-string to this function?
```dart
utf8.decode(buf.toList())
```
Does it returns null? if so, we must have a test case for it.

packages/polkadart_scale_codec/lib/src/core/source.dart
leonardocustodio: I don't see how what is the difference of `compactLenght` and `lengthOfCompact`
They mean exactly the same thing.... Between both I would stay with the first option..

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -201,7 +201,7 @@ class Codec {
 
   /// Decodes Bit Sequence
   List<dynamic> _decodeSequence(SequenceType def, Source source) {
-    int len = source.compactLength();
+    int len = source.lengthOfCompact();
```
leonardocustodio, line 204: No abbreviations... length not len

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -454,7 +454,7 @@ class Codec {
   ///
   /// Returns: `BigInt` | `int`
   dynamic _decodeCompact(CodecCompactType type, Source source) {
-    var n = source.compact();
+    var n = source.uncompact();
```
leonardocustodio, line 457: Also, have my doubts about `uncompact` do we use similar terms to other types?
Shouldn't it be `decode` maybe?

packages/polkadart_scale_codec/test/source_test.dart
```diff
@@ -0,0 +1,278 @@
...
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void unCompactHexFromSource(String compactedHex, dynamic expectedVal) {
+  test(
+      'When a hex:$compactedHex is compacted, it should produce result: $expectedVal.',
+      () {
+    final source = Source(compactedHex);
+
+    var computedVal = source.uncompact();
```
leonardocustodio, line 14: value not val

leonardocustodio (changes requested):

packages/polkadart_scale_codec/lib/src/core/codec.dart
justkawal: As far as I know its doing the opposite of compact.
Otherwise, I'll rename it to `decodeCompact`. Does that sound good ?

packages/polkadart_scale_codec/test/source_test.dart
justkawal: Alright

packages/polkadart_scale_codec/lib/src/core/codec.dart
justkawal: Alright

packages/polkadart_scale_codec/test/source_test.dart
justkawal: Alright, Testing this case scenario.
That's nice to point this thing out.

packages/polkadart_scale_codec/lib/src/core/codec.dart
pastre: The opposite of `compact` is `expand` or `extract`, although in this context I'd prefer `encode`/`decode` or `zip`/`unzip`

justkawal: `extract` sounds good. Thanks @pastre

leonardocustodio (approved):

Issue #52, opened 2022-10-05
Test "Tuple"


justkawal:
Tasks:
- [ ] Test "Tuple"

Acceptance Criteria:
* Edge Cases
* Encode / Decode
* Bad Cases where code breaks:
    * Passing different type to encode/decode function
    * Passing wrong value to encode/decode function

PR #89, opened 2022-10-12
Create tuple type tests

gabrielokura:
## Description
This PR adds tuple type encode and decode tests to the project

## Why?
Issue #52 

## How?
Just adding simple unit tests for mentioned type. 

## Testing Plan 
Run `dart test` on packages/polkadart_scale_codec

### DIFF
--- a/packages/polkadart_scale_codec/test/test_types/tuple_test.dart
+++ b/packages/polkadart_scale_codec/test/test_types/tuple_test.dart
@@ -1,0 +1,173 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  group('encode tuple tests ', () {
+    final typesRegistry = <String, dynamic>{
+      'Codec': {
+        'a': '()',
+        'b': '(Compact<u8>, bool)',
+        'c': '(Compact<u8>, bool, String)',
+      },
+    };
+
+    // Creates the registry for parsing the types
+    final registry = OldTypeRegistry(types: typesRegistry);
+
+    // specifying which type to use
+    registry.select('Codec');
+
+    // fetching the parsed types from `Json` to `Type``
+    final types = registry.getTypes();
+
+    // Initializing Scale-Codec object
+    final codec = Codec(types);
+
+    test(
+        'Given a serie with compact unsigned integer and boolean it should be encoded to correct value',
+        () {
+      const value = [3, true];
+      const expectedResult = '0x0c01';
+
+      final registryIndex = registry.getIndex('(Compact<u8>, bool)');
+
+      expect(codec.encodeToHex(registryIndex, value), expectedResult);
+    });
+
+    test(
+        'Given a serie with compact unsigned integer, boolean and a String it should be encoded to correct value',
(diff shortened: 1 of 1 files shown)


justkawal (changes requested):

packages/polkadart_scale_codec/test/test_types/tuple_test.dart
```diff
@@ -0,0 +1,173 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  group('encode tuple tests ', () {
+    final typesRegistry = <String, dynamic>{
+      'Codec': {
```
justkawal, line 8: I think it's not necessary to initialise `registry` and `Codec` again and again.

justkawal: Initialise it outside of group only once and we're good to use it.

packages/polkadart_scale_codec/test/test_types/tuple_test.dart
```diff
@@ -0,0 +1,173 @@
...
+    registry.select('Codec');
+
+    // fetching the parsed types from `Json` to `Type``
+    final types = registry.getTypes();
+
+    // Initializing Scale-Codec object
+    final codec = Codec(types);
+
+    test(
+        'Given a serie with compact unsigned integer and boolean it should be encoded to correct value',
```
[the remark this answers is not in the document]
gabrielokura: i took it from substrate tuple description. 

"A fixed-size series of values, each with a possibly different but predetermined and fixed type."

But i'll change it to `list`

packages/polkadart_scale_codec/test/test_types/tuple_test.dart
```diff
@@ -0,0 +1,139 @@
...
+void main() {
+  final typesRegistry = <String, dynamic>{
+    'Codec': {
+      'b': '(Compact<u8>, bool)',
+      'c': '(Compact<u8>, bool, String)',
+    },
+  };
+
+  // Creates the registry for parsing the types
+  final registry = OldTypeRegistry(types: typesRegistry);
```
Lohann, line 14: Wondering why we call this `OldTypeRegistry`, what is the diference between this and the regular TypeRegistry?

leonardocustodio: Yep, I don't think we should have `OldTypeRegistry` even if we do use an older registry you should just pass it to the `normal` registry and it should work the same.

leonardocustodio (changes requested):
Why do we only have tests for Compact, bool, and String?
Tuples can have other types as well. Is this PR specific for that?

justkawal:
@gabrielokura 
Can you please add following tests too:
- [ ] `(u8, bool)`
- [ ] `(Text, Option<(u16, u64)>)`

and some more mixes and matches.

justkawal (changes requested):
Can we add following mixed types for tuple:
- [ ] `(Option<u8>, bool)`
- [ ] `(bool, Option<(Text, u8)>)`
- [ ] `(Text, Vec<u8>)`

PR #94, opened 2022-10-16
Remove dynamics on types decoding/encoding

gabrielokura:
## Any follow ups?

Please review #92 before this one.

## Description
This PR adds `generics` and remove `dynamic` values on decode/encode methods. 
Also, removes unnecessary tests.

## Why?
Dynamic values results in a typecast and a bunch of if/else/throw conditions that bring no value to our code.

## How?
1. Add generics to `ScaleCodecType` abstract class.
2. Update each type with its types

## Testing Plan 
Run dart test


Lohann (approved):

justkawal (approved):
LGTM

PR #87, opened 2022-10-11
Add xcm schema

justkawal:
## Any follow ups?
Please review PR: #86 , before reviewing this PR.

## Description
* Adds v0-v2 for xcm.
* Adds `xcm/src.dart` served as and entry point to access xcm schema.

PR #100, opened 2022-10-17
Adds Spec-Version

justkawal:
## Any follow ups?
Review PR: #87, before reviewing this PR.

## Description
* Adds Spec-version-schema
* Adds Spec-Version model class

## Why?
* Will further be helpful to validate **spec of metadata** with this specified **spec-schema (json-schema)** to verify the spec while reading the definitions file.


packages/substrate_metadata/lib/spec_version/spec_version.model.dart
```diff
@@ -0,0 +1,68 @@
+class SpecVersionRecord {
```
gabrielokura, line 1: Could you add some documentation for this class?

Something like you written on description:
"Will further be helpful to validate spec of metadata with this specified spec-schema (json-schema) to verify the spec while reading the definitions file"

leonardocustodio (approved):

gabrielokura (approved):

Issue #113, opened 2022-10-18
Unit Test EventRegistry


justkawal:
Tasks:
- [ ] Unit Test `EventRegistry` from `susbtrate_metadata`.

justkawal:
No more using event registry: #186

Issue #112, opened 2022-10-18
Unit Test TypeHashing


justkawal:
Tasks:
- [ ] Unit Test `TypeHasher` from `susbtrate_metadata`.

PR #114, opened 2022-10-18
Add events, calls, type hashing, exceptions

justkawal:
## Any follow ups?
Please review PR: #100, before reviewing this PR.

## Description
* Adds Exceptions
* Adds Logic for type-hashing (HashNode, TypeHasher):
    - Hashes the list of types and stores their hashed `sha256` in `weak-map`.
    - Helps to do a fast-lookup quickly.
* Adds EventsRegistry logic

## Future task:
* Unit Test `TypeHashing`: #112
* Unit Test `EventRegistry`: #113

## Testing Plan 
* Unit Tests will be added in future PRs  with resolving issues:
    - #112
    - #113


packages/substrate_metadata/pubspec.yaml
```diff
@@ -10,6 +10,12 @@ environment:
   sdk: ">=2.17.6 <3.0.0"
 
 dependencies:
+  cryptography: ^2.0.5 # Apache-2.0
+  polkadart_scale_codec:
+    git:
+      url: https://github.com/rankanizer/polkadart.git
+      path: packages/polkadart_scale_codec
```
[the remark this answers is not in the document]
justkawal: @Lohann Suggestion Committed ✅

packages/substrate_metadata/lib/utils/utils.dart
```diff
@@ -0,0 +1,66 @@
...
+bool isNotEmpty(dynamic value) {
+  if (value == null) {
+    return false;
+  }
+  if (value is bool) {
+    return value;
+  } else if (value is String) {
+    return value.trim() != '';
+  } else if (value is num) {
+    return value != 0;
```
Lohann, line 16: I'm not sure if a number being equal to zero can be considered empty..

packages/substrate_metadata/lib/types_hashing.dart
```diff
@@ -0,0 +1,230 @@
...
+///as we typically have around 10^3 nodes in a graph. This is on a verge of being slow.
+///
+///Hence, the following procedure:
+/// 1. We embed Tarjan's strongly connected components algorithm in our hash computation to discover and
+/// persist information about strongly connected components.
+/// 2. For each strongly connected component, we cache the resulting hash per entry point.
+///
+///This allows us to visit non-mutually recursive types only once and makes the overall procedure
+///quadratic only on the size of a maximal strongly connected component, which practically should not be big.
+///
```
Lohann, line 64:
```suggestion
///
/// Computes hashes of substrate types for the purpose of equality derivation.
///
/// Substrate types form a cyclic directed graph.
/// Two types are equal when their depth-first traversal trees are equal.
/// Inline with equality, we define a type hash to be a merkel hash of it's depth-first traversal.
///
/// Note, that unlike a classic tree case we might need
/// to visit mutually recursive type nodes more than once.
///
/// Naive approach of performing a depth-first traversal for each node might not work,
/// as we typically have around 10^3 nodes in a graph. This is on a verge of being slow.
///
/// Hence, the following procedure:
/// 1. We embed Tarjan's strongly connected components algorithm in our hash computation to discover and
/// persist information about strongly connected components.
///  2. For each strongly connected component, we cache the resulting hash per entry point.
///
/// This allows us to visit non-mutually recursive types only once and makes the overall procedure
/// quadratic only on the size of a maximal strongly connected component, which practically should not be big.
///
```

packages/substrate_metadata/lib/types_hashing.dart
```diff
@@ -0,0 +1,230 @@
...
+  var hasher = HASHERS.get(types);
+  if (hasher == null) {
+    hasher = TypeHasher(types);
+    HASHERS[types] = hasher;
+  }
+  return hasher;
+}
+
+///
+///Get a strong hash of substrate type, which can be used for equality derivation
```
Lohann, line 20:
```suggestion
/// Get a strong hash of substrate type, which can be used for equality derivation
```

Lohann (changes requested):

Lohann (approved):

PR #115, opened 2022-10-18
Adds mapped models for metadata

justkawal:
# Why big PR:
* Model Classes are calling each other and are internally tightly couple with each other.
* Its just an image of the metadata of the chain.
* Why tightly coupled: **( Metadata reuses the available types with the use of mapped keys. )**

## Any follow ups?
Please review PR: #114 , before reviewing this PR.

## Description
* This PR adds model classes for the substrate type bundles.
* Model classes is a 1-1 mapping to be filled with the decoded metadata of the chains.

## How helpful ?
* When we will decode the chain in the future PRs, then it will help us to traverse through the metadata with types available on the go without having to look for keys in `hashmaps` and then casting the types again and again.
* It helps in parsing the `ChainDescription` of the specified chain.

### DIFF
--- a/.github/workflows/reuseable_workflow.yml
+++ b/.github/workflows/reuseable_workflow.yml
@@ -54,7 +54,7 @@ jobs:
         run: dart pub global run dart_code_metrics:metrics analyze --reporter=codeclimate .
 
       - name: Test & publish code coverage
-        uses: paambaati/codeclimate-action@v3.1.0
+        uses: paambaati/codeclimate-action@v3.1.1
         env:
           CC_TEST_REPORTER_ID: f62061f125a8fc1425670a787b7cac15d2d9f73d7ec4878c0e4d07575156ccc1
         with:


packages/substrate_metadata/lib/models/storage/storage.dart
```diff
@@ -0,0 +1,246 @@
...
+      );
+}
+
+///
+/// -------------- Storage Entry [Ends] --------------
+///
+
+///
+/// -------------- Storage Hasher [Starts] --------------
+///
```
Lohann, line 92: **[optional suggestion]**
Why not split this into two files?

packages/substrate_metadata/lib/models/metadata/metadata_base.dart
```diff
@@ -0,0 +1,25 @@
+part of models;
+
+class Metadata {
+  final String kind;
+  final dynamic value;
+  const Metadata({required this.kind, required this.value});
+
+  /// Creates Metadata Object from the version info
+  static Metadata fromVersion(Map<String, dynamic> json, int version) {
```
Lohann, line 9: I don't see tests for this method, is this tested already?

packages/substrate_metadata/lib/models/version_description.model.dart
```diff
@@ -0,0 +1,19 @@
+part of models;
+
+class VersionDescription extends SpecVersion {
```
Lohann, line 3: Where we use it?

Lohann (changes requested):
I see a lot of new code, but no examples on how to use it, or tests.

packages/substrate_metadata/lib/models/storage/storage.dart
justkawal: Done ✅

leonardocustodio (approved):

leonardocustodio (approved):

leonardocustodio (approved):

gabrielokura (approved):

PR #121, opened 2022-10-19
Metadata decoder

justkawal:
## Any follow ups?
Please review PR: #115, before reviewing this PR.

## Description
* Adds Metadata Decoder.
* Added `substrate_metadata` to `test.yml`. (To include tests from substrate_metadata in CI)

## Test
`dart test`


leonardocustodio (approved):

packages/substrate_metadata/lib/metadata_decoder.dart
```diff
@@ -0,0 +1,118 @@
+import 'dart:typed_data';
+import 'models/models.dart' as model;
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart' as scale;
+import 'package:substrate_metadata/old/definitions/metadata/metadata.dart'
+    as metadata_definitions;
+
+/// (Singleton)
```
pastre, line 7: Why is this a singleton?

justkawal: Because then it would process all the metadata definitions again and again by calling _createScaleCodec() which is a heavy task.

So avoiding the same repetition of this task.

packages/substrate_metadata/lib/metadata_decoder.dart
```diff
@@ -0,0 +1,118 @@
...
+///
+/// It helps to create a single instance of [MetadataDecoder] which helps to decode Metadata.
+class MetadataDecoder {
+  static final MetadataDecoder instance = MetadataDecoder._internal();
```
Lohann, line 11: Singleton is an anti-pattern in modern applications and must be used only in very specific scenarios. A singleton is shared mutable state, and usually makes the app really hard to test (because all tests touch the global state).

Singleton can be replaced by `Context` objects, which instead of having a global reference, we pass then as parameters to parts of the application which needs a shared state.

Lohann (changes requested):

packages/substrate_metadata/lib/metadata_decoder.dart
Lohann: https://codewithandrea.com/articles/flutter-singletons/

justkawal: Resolved ✅

packages/substrate_metadata/lib/metadata_decoder.dart
```diff
@@ -0,0 +1,118 @@
...
+  ///
+  /// Map<String, dynamic> decodedMetadataMap = decoderInstance.decode('0x090820....');
+  /// ```
+  Map<String, dynamic> decode(dynamic data) {
+    final result = _decodePrivate(data);
+    return result['metadata'];
+  }
+
+  /// Private method to decode metadata
+  Map<String, dynamic> _decodePrivate(dynamic data) {
```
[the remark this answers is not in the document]
justkawal: It is okay, Try to be little bit flexible, buddy.

packages/substrate_metadata/lib/metadata_decoder.dart
```diff
@@ -0,0 +1,118 @@
...
+    }
+    var source = scale.Source(content);
+
+    var magic = source.u32();
+    scale.assertionCheck(
+        magic == 0x6174656d, 'No magic number 0x6174656d at the start of data');
+
+    var version = source.u8();
+    scale.assertionCheck(
+        9 <= version && version < 15, 'Invalid metadata version');
```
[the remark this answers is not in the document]
justkawal: Resolved 🤣

packages/substrate_metadata/lib/metadata_decoder.dart
```diff
@@ -0,0 +1,118 @@
...
+    if (data is String) {
+      content = scale.decodeHex(data);
+    } else {
+      content = data;
+    }
+    var source = scale.Source(content);
+
+    var magic = source.u32();
+    scale.assertionCheck(
+        magic == 0x6174656d, 'No magic number 0x6174656d at the start of data');
```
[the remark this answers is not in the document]
justkawal: Resolved 🤣

justkawal:
Resolved as commented by Bruno.

packages/substrate_metadata/lib/metadata_decoder.dart
justkawal: > I've also seen people use a `Singleton+`, were objects have a shared instance and a public constructor. As long as its abstraction is done right and it is properly injected, it is safe to use. This is not the case of this class, just sharing knowledge.
> 
> Also: why do we need to share the decoder? Isn't it stateless?

Bruno sometimes, its important to run before starting to walk.

PR #140, opened 2022-10-20
Create new codec type for Compact

gabrielokura:
## Description
This PR adds a new codec type for encode/decode Compact integers.

## Why?
Refactoring `Codec`. 

## How?
1. Create new CodecCompact class
2. Create its tests for encoding/decoding values

## Testing Plan 
Run `dart test` in polkadart_scale_codec package folder

### DIFF
--- a/packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
+++ b/packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
@@ -1,0 +1,98 @@
+part of '../core/core.dart';
+
+/// [ScaleCodecType] class to encode and decode `signed 8-bit` integers.
+///
+/// A "compact" or general integer encoding is sufficient for encoding
+/// large integers (up to 2**536) and is more efficient at encoding most
+/// values than the fixed-width version. (Though for single-byte values,
+/// the fixed-width integer is never worse.)
+///
+/// See also: https://docs.substrate.io/reference/scale-codec/
+class CodecCompact<T extends Object> implements ScaleCodecType<T> {
+  /// Default constructor used to limit [T] types.
+  /// Only `int` and `BigInt` are supported.
+  CodecCompact() {
+    if ((T != int && T != BigInt)) {
+      throw UnexpectedTypeException(
+          expectedType: 'int or BigInt', receivedType: T.toString());
+    }
+  }
+
+  /// Returns an encoded hex-decimal `string` from `int` or `BigInt`[value]
+  /// using [HexEncoder] methods.
+  ///
+  /// Compact/general integers are encoded with the two least significant bits
+  /// denoting the mode:
+  /// - 0b00: single-byte mode; upper six bits are the LE encoding of the value
+  ///  (valid only for values of `0-63`).
+  /// - 0b01: two-byte mode: upper six bits and the following byte is the LE
+  ///   encoding of the value (valid only for values `64-(2**14-1)`).
+  /// - 0b10: four-byte mode: upper six bits and the following three bytes are
+  ///   the LE encoding of the value (valid only for values `(2**14)-(2**30-1)`).
+  /// - 0b11: Big-integer mode: The upper six bits are the number of bytes following,
+  ///   plus four. The value is contained, LE encoded, in the bytes following.
+  ///   The final (most significant) byte must be non-zero. Valid only for values
+  ///   `(2**30)-(2**536-1)`.
+  ///
+  /// Example:
+  /// ```
+  /// final intEncoded = CodecCompact<int>().encodeToHex(65535); //"0xfeff0300"
+  /// final bigIntEncoded = CodecCompact<BigInt>().encodeToHex(18446744073709551615.toBigInt);
+  /// //0x13ffffffffffffffff
+  /// ```
+  @override
+  String encodeToHex(value) {
+    var sink = HexEncoder();
+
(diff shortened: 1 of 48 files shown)


gabrielokura:
@pastre If you could check this out. I think your insights would be much helpful in this different approach

packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
```diff
@@ -0,0 +1,98 @@
...
+  /// ```
+  /// final intEncoded = CodecCompact<int>().encodeToHex(65535); //"0xfeff0300"
+  /// final bigIntEncoded = CodecCompact<BigInt>().encodeToHex(18446744073709551615.toBigInt);
+  /// //0x13ffffffffffffffff
+  /// ```
+  @override
+  String encodeToHex(value) {
+    var sink = HexEncoder();
+
+    var convertedValue = value is int ? value : (value as BigInt).toInt();
```
pastre, line 47: `(value as BigInt).toInt()` might be a bad idea, as values larger than what `int` supports will be clamped Look for an approach where you don't need to convert types, nor check for them in real time. A simple approach would be to treat everything as a bigint

pastre (changes requested):

pastre:
Generics is a tool that allows us to not specify types while keeping all the benefits from string typing. If we're gonna type cast and throw, it might be simpler to just keep stuff as simple as possible

packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
gabrielokura: Done! Changed CodecCompact to support only BigInt and let the user manipulates the result and transform to int (if wants)

packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
```diff
@@ -0,0 +1,82 @@
...
+  /// Example:
+  /// ```
+  /// String example1 = CodecCompact().encodeToHex('65535'.toBigInt); //"0xfeff0300"
+  /// String example2 = CodecCompact().encodeToHex(18446744073709551615.toBigInt);
+  /// //0x13ffffffffffffffff
+  /// ```
+  @override
+  String encodeToHex(value) {
+    var sink = HexEncoder();
+    int convertedValue = value.toInt();
```
pastre, line 37: Conversion to int might be clamped. Why aren't we using bigint?

pastre: If you need it to be an int, consider creating a guard clause for using  `value.isValidInt`. This will mitigate the pyramid of doom that's happening down here

packages/polkadart_scale_codec/lib/src/types/codec_compact.dart
```diff
@@ -0,0 +1,82 @@
...
+    if (convertedValue < 64) {
+      sink.write(convertedValue * 4);
+    } else if (convertedValue < pow(2, 14).toInt()) {
+      sink.write((convertedValue & 63) * 4 + 1);
+      sink.write(convertedValue >>> 6);
+    } else if (convertedValue < pow(2, 30).toInt()) {
+      sink.write((convertedValue & 63) * 4 + 2);
+      sink.write((convertedValue >>> 6) & 0xff);
+      sink._uncheckedU16(convertedValue >>> 14);
+    } else if (convertedValue < calculateBigIntPow(2, 536).toInt()) {
```
pastre, line 52: `calculateBigIntPow(2, 536)` doesn't fit in an int (dart ints are as large as 2^53) and is getting clamped to 2^53. Really unsure if this is working

packages/polkadart_scale_codec/test/test_types/codec_compact_test.dart
```diff
@@ -0,0 +1,104 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('encode compact integers to hex', () {
+    test('Should return correct encoded data when value is zero', () {
+      final value = 0.toBigInt;
```
pastre, line 7: nit: BigInt has an API for 0, 1 and 2 (chicken count lol)
```suggestion
      final value = BigInt.zero;
```

packages/polkadart_scale_codec/test/test_types/codec_compact_test.dart
```diff
@@ -0,0 +1,104 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('encode compact integers to hex', () {
```
pastre, line 5: Test cases should be:
- [X] Negative int should throw
- [X] 0  should encode on first condition
- [ ] 63 should encode on first condition
- [ ] 64 should encode on second condition
- [ ] 2^14 - 1 should encode on second condition
- [X] 2^14 should encode on third condition
- [ ] 2^30 - 1 should encode on third condition
- [X] 2^30 should encode on last condition
- [ ] 2^536 - 1 should encode on last condition
- [ ] 2^536 should throw

I'm happy with an issue to refactor

pastre: It's way more valuable to test the boundaries instead of a random value (like 69). If somebody was to change from 64 to 65, your tests would not fail

PR #141, opened 2022-10-21
Test extrinsics

justkawal:
## Any follow ups?
Please review PR: #136, before reviewing this PR.

## Description
* Few bug fixes.
* Few important refactoring and functionality improvement.
* Adds Extrinsic Test for `polkadot` chain.
* Adds polkadot `chain definition`.

## Testing Plan
`dart test`

PR #142, opened 2022-10-21
Test Events

justkawal:
## Any follow ups?
Please review PR: #141, before reviewing this PR.

## Description
* Adds events testing

## Testing Plan 
`dart test`


packages/substrate_metadata/test/events_test.dart
```diff
@@ -0,0 +1,75 @@
...
+    //
+    // Populating with the metadata for block-numbers available for this chain....
+    chain.initSpecVersionFromFile('../../chain/polkadot/versions.jsonl');
+
+    final List<RawBlockEvents> rawBlocksList =
+        getEvents('../../chain/polkadot/events.jsonl');
+
+    //
+    // Looping through every block
+    for (var originalEvent in rawBlocksList) {
```
gabrielokura, line 51: these tests don't follow last bruno's tips.

He said we should avoid loops and must create readable tests titles

justkawal: I have a call with him, and for this case scenario we're going with it by like these.

justkawal: Its practically not possible, to search a sweet block which would be good enough to test. So, can't cherry pick any one.

gabrielokura:
I see we are testing events, but what is the meaning of these tests? How they protect us against bugs in decode && encode methods?

I mean, shouldn't we test these two different methods separately?

justkawal:
> I see we are testing events, but what is the meaning of these tests? How they protect us against bugs in decode && encode methods?
> 
> shouldn't we test these two different methods separately?

When you decode a value with use of the metadata definition only and then encode it back and get the same result as the original value then it confirms that the code understands the inputed value, as resulting in E2E testing....

Again, I'm not saying the code is 100% sweet and tight, but we will get it in future with `unit-tests.`

Future task: `unit-test` @gabrielokura .

gabrielokura:
> 

understood!

gabrielokura:
> > I see we are testing events, but what is the meaning of these tests? How they protect us against bugs in decode && encode methods?
> > shouldn't we test these two different methods separately?
> 
> When you decode a value with use of the metadata definition knowledge only and then encode it back and gets the same result as the original value then it confirms that the code understands the inputed value with the use of provided definitions or schema or metadata, as resulting in E2E testing....
> 
> Again, I'm not saying the code is 100% sweet and tight, but we will get it in future with `unit-tests.`
> 
> Future task: `unit-test` @gabrielokura .

understood!

packages/substrate_metadata/test/events_test.dart
```diff
@@ -0,0 +1,77 @@
...
```
pastre, line 51: Please create an issue to track and refactor this

pastre (changes requested):
Need to track bad tests

packages/substrate_metadata/test/events_test.dart
justkawal: Tracked with #150.

leonardocustodio (approved):

pastre (approved):

PR #166, opened 2022-12-16
Test Cases with Optimisations

justkawal:
## Description
- Test-Cases with manual Metadata usage:
    - `Extrinsic`
    - `Events`
    - `Constants`
- Bug Fixes
- Optimised by pre-checking `isPreV14` in `Chain()` Wrapper.

## Why?
As asked in the review to show an example / test-case where direct / manual usage of Metadata is demonstrated without using LegacyTypes.

## How?
- By decoding Metadata - V14, and then creating ChainDescription from it, we could use the` ChainDescription` nearly in everything without even knowing the chain-definition.
- In this way `LegacyTypes` is not needed when used `Metadata - V14`.

## Testing Plan 
`dart test`


justkawal:
@Lohann Can you review it please ?

Most of it is just parsed data for `events and extrinsics test-cases` and actual understandable code is less than `200 lines`.

leonardocustodio (approved):

packages/polkadart_scale_codec/lib/src/core/registry.dart
```diff
@@ -80,6 +82,44 @@ class TypeRegistry {
     return types;
   }
 
+  List<Type> _fixU256Structs(List<Type> types) {
```
Lohann, line 85: Did you created this fix? or is it from subsquid? can you put a reference why we need this fix?

Lohann (changes requested):

packages/polkadart_scale_codec/lib/src/core/registry.dart
Lohann: this seems a bit hack

justkawal: Alright will put a reference in code tomorrow morning:

https://github.com/subsquid/squid-sdk/blob/5ece47edd3bd8da88643cc6a735b13eea2a5eeb6/substrate/substrate-metadata/src/util.ts#L46

Lohann (approved):

packages/polkadart_scale_codec/lib/src/core/registry.dart
Lohann: Aweosme, thanks for the clarification @justkawal !! approved

PR #172, opened 2022-12-21
Remove dart mirrors

### LINKED ISSUE #170: Remove dart:mirrors
- Remove dart:mirrors from polkadart_scale_codec

justkawal:
## Description
- Fix #170 

## Why?
- Helps to make `exe` and `js` executables when `dart:mirrors` is removed.
- Helps to improve performance to know `ahead of time` function call.
- Helps to avoid further floating of code. (The reason why `dart:mirrors` is still limited to dart only but not available to flutter frame-work.)

## Testing Plan 
`flutter test`


leonardocustodio (approved):

pastre (approved):
Nice! Mirrors is also a very heavy lib, which can bloat build results size to about 5 times!

Issue #176, opened 2022-12-26
Conversion - Account Address


justkawal:
- Convert Account Address to `hex` format from buffer bytes.
- Make separate decoding method to handle decoding of address.

Issue #169, opened 2022-12-21
Add toJson() functionality for extrinsics and events


justkawal:
- Add a method `toJson()` functionality.

PR #177, opened 2022-12-26
Change decoding bytes array

justkawal:
## Description
- Add different way to decode bytes buffer to return account address.
- Add `toJson()`.

## Why?
- Resolves Issue: #176 
- Resolves Issue: #169


packages/substrate_metadata/lib/utils/utils.dart
```diff
@@ -67,3 +67,34 @@ String sha256(dynamic data) {
...
+extension ToJson<T> on T {
+  T toJson() {
+    if (this is List<T>) {
+      return (this as List<T>).map((T e) => e.toJson()).toList() as T;
+    }
+    if (this is List<Map<String, dynamic>> || this is Map<String, dynamic>) {
+      return jsonDecode(toHuman());
+    }
+    return this;
+  }
```
pastre, line 81: I didn't get this function: does it convert from json or to json? It doesn't look like we get a string from this. If we get a JSON object from it (dictionary, list or type), I'd rename it to `makeJsonObject()` to avoid misunderstanding the point of this function

packages/substrate_metadata/lib/utils/utils.dart
```diff
@@ -67,3 +67,34 @@ String sha256(dynamic data) {
...
+
+  String toHuman() {
```
pastre, line 83: same as above: if it returns a serialized object in form of a string, we should name it accordingly: `toJson()` is appropriate here

Another suggestion for both names is `serialize` and `deserialize`: it makes it clear that we're creating a JSON string or creating a JSON object

packages/substrate_metadata/lib/utils/utils.dart
```diff
@@ -67,3 +67,34 @@ String sha256(dynamic data) {
...
+      return jsonEncode(
+        this,
+        // handle BigInt and Some
+        toEncodable: (value) {
+          if (value is BigInt) {
+            return value.toString();
+          } else if (value is scale.Some) {
+            return value.toString();
+          }
+          return value;
```
pastre, line 94: Can't we have just `value.toString()`? Also, can't we do this in a single statement
```suggestion
          if (value is BigInt || value is scale.Some) {
            return value.toString();
          }
          return value;
```

pastre (changes requested):

packages/substrate_metadata/lib/utils/utils.dart
gabrielokura: Could be :
```suggestion
          return (value is BigInt || value is scale.Some) ? value.toString() : value;
```

packages/substrate_metadata/lib/utils/utils.dart
leonardocustodio: This function should return a JSON object as String. Pretty much like `JSON.stringify`
Imo, we should keep as `.toJson()` as this name is used in a lot of substrate libs. Like the `toHuman`.

leonardocustodio: I believe we should be consistency with a few basic types from the Polkadot-JS API. As they are used across a lot of substrate libs. We can find the description about those in here: https://polkadot.js.org/docs/api/start/types.basics/
So I would keep it as `.toJSON` as well as the other names and if we don't have all of them probably make a few more extensions.

packages/substrate_metadata/lib/utils/utils.dart
leonardocustodio: @justkawal can you give me an example of a extrinsic with Some using this toJson? I want to check how it appears

justkawal: It will appear as: `"Some(Some(123))"` or `"Some(None)"`.

Here: https://github.com/rankanizer/polkadart/blob/930fec67fa737bb234e90ffac8f8c99d70bac2c1/packages/polkadart_scale_codec/lib/src/core/option_type.dart#L51

but I'm planning to change that and remove word `Some` from the string in the upcoming PR.

leonardocustodio: Ok not good. In my opinion we should make it just like an enum so it should be something like this:
`{some_key: {None: null}}}`or `{some_key: {Some: 10}}` or `{some_key: {Some: null}}`

leonardocustodio: Obviously you need to check recursively as we could have Some<Some<Some<T>>>

justkawal: So after removing word it would be evaluated like:

```dart
@override
String get toString() {
    if (value is _Option) {
        return (value as _Option).value.toString();
    }
    return value.toString();
}
```

So: `Some(Some(Some(123)))` would evaluated to `"123"`
So: `Some(Some(None))` would evaluated to `"None"`

leonardocustodio: It is important to know if it was Some and not just the final value. For example when decoding extrinsic, I want to know the exact arg.

justkawal: Like: `Some(Some(Some(123)))`  ->  `{ "some_key": { "Some": { "Some": { "Some": 123 } } } }` ?

leonardocustodio: Yep, I believe that's the way to go, what do you think @Lohann

pastre: It might get confused with a dictionary/custom data structure. It's fine for encoding, because you know the type you're encoding to, but you couldn't differ nested optional values from a dictionary while decoding. Most solutions solve this reserving a keyword and documenting that this conflict might happen if the user tries to use reserved keys while decoding.

packages/substrate_metadata/lib/utils/utils.dart
pastre: Hm, not sure I agree 😅  I think we should find the best solution for our context, which is very different from JS. But no reason to block a PR on this :)

packages/substrate_metadata/lib/utils/utils.dart
pastre: Real solution for this would be to refactor everything to make it strongly typed, but it's probably too late for that

leonardocustodio: That's for the "json" output. I believe we do have the types when using default dart objects, don't we?
But yeah in theory it could be a bit confusing if using the reserved keys, for example, decoding the String Some: `{some_key: "Some"}` though it is important to know the context here, that to decode anything you need to know the metadata. So you would already know that's an enum and not a string. Because otherwise you can't even decode it.

justkawal: Exactly, also I don’t see a future problem arising from it.

packages/substrate_metadata/lib/utils/utils.dart
justkawal: It’s common everyday term -> function name we use and understands easily.

justkawal: If we would peek into documentation by Big Tech companies then we would find the similar documentation and function names….

Also using `toJson()` seems normal and good fit to me.

I would prefer to stick with that.

packages/substrate_metadata/lib/utils/utils.dart
justkawal: @leonardocustodio  I would also give something like: `Some.fromJson(Map<String, dynamic> map)` which would help to create the Some or None Object back.

justkawal:
@leonardocustodio Can you please approve this PR ?

leonardocustodio:
Not really, I see you took the idea of the squid approach. But there is a reason why kind is underlined in there. So we know it is not a struct. You should do the same if you are going that way.

leonardocustodio (approved):

justkawal:
@Lohann Can you please review the request ?

leonardocustodio (approved):

packages/polkadart_scale_codec/lib/src/core/option_type.dart
```diff
@@ -50,4 +64,44 @@ class Some<T> extends _Option implements EquatableMixin {
...
+  ///
+  /// Some(None).toJson() => {'_kind': 'Some', 'value': {'_kind': 'None'}}
+  /// ```
+  Map<String, dynamic> toJson() {
+    return {
+      '_kind': kind,
+      'value': value is Some
+          ? (value as Some).toJson()
+          : (value == None ? None.toJson() : value),
+    };
```
pastre, line 86: I'd avoid nesting ternaries

packages/substrate_metadata/lib/utils/utils.dart
pastre: Just FYI

cJSON uses encode/decode
python uses json.loads/dumps
Swift uses JSONEncoder/Decoder
serde uses from_string/to_string
Kotlin uses JSON.stringify/parse

Even native JS uses JSON.stringify/object... The only reference I know that uses `toJSON()` and `toHuman()` is that JS library

Just sharing, not asking for a change :)

pastre (approved):

packages/polkadart_scale_codec/lib/src/core/option_type.dart
justkawal: why so ?

Issue #179, opened 2022-12-28
Fix Bit-Sequence


justkawal:
- Fix Encode
- Fix Decode

justkawal:
Completed functionality fixed and made by @Lohann in #186

PR #180, opened 2022-12-28
Fix bit sequence

justkawal:
## Any follow ups?
#177 

## Description
- Fix Bit Sequence Decoding and Encoding

## Why?
#179 

## Testing Plan 
`dart test`


packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -454,17 +454,35 @@ class Codec {
...
-    return source.bytes(length).toList();
+    final int length = source.compactLength();
+    final List<int> bits = source.bytes((length / 8).ceil()).toList();
+
+    final List<int> result = List<int>.filled(length, 0);
+
+    for (var i = 0; i < length; i++) {
+      final index = (i / 8).floor();
+      result[i] = (bits[index] >> i) & 1;
+    }
```
Lohann, line 465: There's a bug here when the `length > 8`, please create a test for this.

Lohann: Also what you are doing here is exactly what dart's [bit_array](https://pub.dev/packages/bit_array) does in a more performatic way.

https://pub.dev/documentation/bit_array/latest/bit_array/BitArray-class.html

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -454,17 +454,35 @@ class Codec {
...
+        'BitSequence can have bits of type `List<int>` only.');
+
+    encoder.compact(bits.length);
+
+    final List<int> bytes = List<int>.filled((bits.length / 8).ceil(), 0);
+
+    for (var i = 0; i < bits.length; i++) {
+      final index = (i / 8).floor();
+      bytes[index] |= (bits[i] << i);
+    }
```
Lohann, line 483: Same case here, it doesn't work when `length > 8`

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
```diff
@@ -11,38 +11,128 @@ void main() {
...
+      final decoded = Codec(types).decode(usageIndex, '0x201b');
+      expect([1, 1, 0, 1, 1, 0, 0, 0], equals(decoded));
+    });
+
+    test(
+        'When \'0x20ff\' is decoded then result \'[1, 1, 1, 1, 1, 1, 1, 1]\' is produced',
+        () {
+      final decoded = Codec(types).decode(usageIndex, '0x20ff');
+      expect([1, 1, 1, 1, 1, 1, 1, 1], equals(decoded));
+    });
```
Lohann, line 101: Please include more tests for with array greater than 8 bits, also use the rust code as reference.

Lohann (changes requested):

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
```diff
@@ -2,123 +2,195 @@ import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
...
-  final usageIndex = registry.getIndex('BitVec<u8>');
+    // specifying which type to use.
+    final usageIndex = registry.getIndex('BitVec<u8, bitvec::order::Lsb0>');
 
-  // fetching the parsed types from `Json` to `Type`
-  final types = registry.getTypes();
+    // fetching the parsed types from `Json` to `Type`
+    final types = registry.getTypes();
+    test(
+        'When \'[1, 1, 0, 1, 1]\' is encoded then result \'0x141b\' is produced',
```
pastre, line 17: We need meaningful test messages for all tests

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
```diff
@@ -2,123 +2,195 @@ import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
...
+    // Exception on Encode at (lowest - 1) and (highest + 1): `BitVec<u8, bitvec::order::Lsb0>`
+    group('Exception on Encode BitVec<u8, bitvec::order::Lsb0>:', () {
+      test(
+          'should throw exception \'UnexpectedCaseException\' when encoding [-1] which is 1 lesser than lowest acceptable.',
+          () {
+        expect(
+            () => Codec(types).encode(usageIndex, [-1]),
+            throwsA(predicate((e) =>
+                e is UnexpectedCaseException &&
+                e.toString() == 'Invalid byte, unable to encode [-1].')));
```
pastre, line 130: We need a single failing condition for each expectation

pastre: This also applies to all tests

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,80 @@ class Codec {
...
-    encoder.compact(bits.length * 8);
-    encoder.bytes(bits);
+  void _encodeBitSequence(
+      BitSequenceType def, dynamic bits, ScaleCodecEncoder encoder) {
+    assertionCheck(bits is List<int>,
+        'BitSequence can have bits of type `List<int>` only.');
+
+    int capacity = max(bits.length, _getBitSequenceCapacity(def.bitStoreType));
+
+    capacity = (capacity / 8).ceil() * 8;
```
pastre, line 494: `(capacity / 8).ceil()` is repeating a few times, might be a good idea to create a variable for it

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,80 @@ class Codec {
...
+  int _getBitSequenceCapacity(int bitStoreTypeIndex) {
+    // setting default capacity to 8
+    if (_types[bitStoreTypeIndex].kind != TypeKind.Primitive) {
+      return 8;
+    }
+    final type = _types[bitStoreTypeIndex] as PrimitiveType;
+    final name = type.primitive.name.toLowerCase();
+    if (name.startsWith('u') == false) {
+      throw UnexpectedCaseException(
+          'BitVec can have Primitives types of unsigned bit only.');
```
pastre, line 526: nit: improve message
```suggestion
      throw UnexpectedCaseException(
          'Expected unsigned bit with `u` prefix, but got $name which can't be decoded');
```

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,80 @@ class Codec {
...
+    }
+
+    return int.parse(name.substring(1));
```
pastre, line 529: What if `name` is `u`?

pastre (changes requested):

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
justkawal: @pastre Like ?

packages/polkadart_scale_codec/lib/src/core/core.dart
```diff
@@ -5,6 +5,7 @@ import 'dart:math';
 import 'dart:core' as core;
 import 'dart:core';
 import 'dart:typed_data';
+import 'package:bit_array/bit_array.dart';
```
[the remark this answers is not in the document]
justkawal: Yes, it is. That's why its imported here!! I guess.... 👀

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
justkawal: I want to be sure that the error is showing the correct error message and its not mistakenly edited in ay further PR. Thats why I prefer to keep the error messages check also.

packages/polkadart_scale_codec/lib/src/core/codec.dart
justkawal: The operation is being done on local variable and I guess there is no need to create the variable or method outside.

packages/polkadart_scale_codec/lib/src/core/registry.dart
```diff
@@ -526,12 +519,36 @@ class TypeRegistry {
...
+              final int orderTypeIndex = _use(resgitryList[1]);
+              return BitSequenceType(
+                bitStoreType: storeTypeIndex,
+                bitOrderType: orderTypeIndex,
+              );
+            default:
+              throw UnexpectedCaseException(
+                  'Unexpected BitVec initialization. Accepted: BitVec, BitVec<u8, bitvec::order::Lsb0>, BitVec<u32, bitvec::order::Lsb0>');
+          }
+        }
```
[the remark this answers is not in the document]
justkawal: Done

packages/polkadart_scale_codec/lib/src/core/registry.dart
```diff
@@ -526,12 +519,36 @@ class TypeRegistry {
...
+                bitStoreType: _use('U8'),
+                bitOrderType: _use('bitvec::order::Lsb0'),
+              );
+            case 2:
+              final resgitryList = assertTwoParams(type);
+              final bit = resgitryList[0];
+              assertionCheck(
+                  bit is RegistryNamedType &&
+                      bit.name.toLowerCase().startsWith('u'),
+                  'BitVec first param must be u8, u16, u32, u64, u128 or u256');
```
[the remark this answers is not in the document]
justkawal: Done

packages/polkadart_scale_codec/lib/src/core/codec.dart
justkawal: Catched and checked.

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,84 @@ class Codec {
...
+
+    final BitArray bitArray = BitArray.fromUint8List(Uint8List.fromList(bytes));
+
+    final List<int> bits = List.filled(length, 0);
+
+    for (var i = 0; i < length; i++) {
+      bits[i] = bitArray[i] ? 1 : 0;
+    }
+
+    return bits;
```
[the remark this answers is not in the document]
Lohann: The logic here can be simplified as following:

```dart
// get compacted length for filling result
final int length = source.compactLength();

// get the encoded bytes from the source
final Uint8List bytes = source.bytes((capacity / 8).ceil());

// Make bytes.length multiple of 4
final int extraZeros = (4 - (bytes.length % 4)) % 4;
for (var i = 0; i < extraZeros; i++) {
   bytes.insert(0, 0);
}

// Convert Uint8List to BitArray
final bits = BitArray.fromUint8List(bytes);

return bits;
```

Obs: it only works for `LSB`, for `MSB` more code and conversion is needed.

packages/polkadart_scale_codec/test/test_types/bitvec_u32_test.dart
```diff
@@ -0,0 +1,196 @@
...
+      final decoded = Codec(types).decode(usageIndex, '0x201b000000');
+      expect([1, 1, 0, 1, 1, 0, 0, 0], equals(decoded));
+    });
+
+    test(
+        'When \'0x4000000000\' is decoded then result \'[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]\' is produced',
+        () {
+      final decoded = Codec(types).decode(usageIndex, '0x4000000000');
+      expect([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], equals(decoded));
+    });
```
Lohann, line 109: Also create a test with non-zero array, this is a good one to test because is assymetric:
```
[1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 1]
```

Lohann (changes requested):

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,84 @@ class Codec {
...
+    if (name.startsWith('u') == false) {
+      throw UnexpectedCaseException(
+          'Expected unsigned bit with `u` prefix, but got $name which can\'t be decoded');
+    }
+    if (name.length == 1 || int.parse(name.substring(1)) % 8 != 0) {
+      throw UnexpectedCaseException(
+          'Expected unsigned bit with valid bitSize, but got ${name.length == 1 ? '' : name.substring(1)} which can\'t be decoded.');
+    }
+
+    return int.parse(name.substring(1));
```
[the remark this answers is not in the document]
justkawal: Seems a good one 💯. Nice, Thanks, Done

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,84 @@ class Codec {
...
+
+    encoder.bytes(
+        bitArray.byteBuffer.asUint8List(0, (capacity / 8).ceil()).toList());
+  }
+
+  int _getBitSequenceCapacity(int bitStoreTypeIndex) {
+    // setting default capacity to 8
+    if (_types[bitStoreTypeIndex].kind != TypeKind.Primitive) {
+      return 8;
+    }
```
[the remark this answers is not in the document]
justkawal: Done ✅

packages/polkadart_scale_codec/lib/src/core/codec.dart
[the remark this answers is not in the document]
justkawal: An issue to track `MSB` is already created: #181

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
pastre: I agree that we should keep both checks, but I think it's important that we're clear on the error message that we display in our framework. We want developers to be sure of what's going wrong, therefore we need to be specific and as detailed as possible when something fails 
```suggestion
            throwsA(predicate((e) {
                expect(e, isA<UnexpectedCaseException>());
                expect(e.toString(), 'Invalid byte, unable to encode [-1].');
            }
```

In the above suggestion, it's clear why our test fails

packages/polkadart_scale_codec/test/test_types/bitvec_test.dart
pastre: Why are you testing that input? What functionality that test protects? Why is that output produced?
I can't extract meaning out of your current description

packages/polkadart_scale_codec/lib/src/core/codec.dart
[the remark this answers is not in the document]
Lohann: Did you push the changes?  I don't see the refactor

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,84 @@ class Codec {
...
+          break;
+        default:
+          throw UnexpectedCaseException(
+              'Invalid byte, unable to encode $bits.');
+      }
+    }
+
+    encoder.bytes(
+        bitArray.byteBuffer.asUint8List(0, (capacity / 8).ceil()).toList());
+  }
```
pastre, line 515: @justkawal this is what I mean by create a new variable - also doing some clean up
```suggestion
    final int capacity = max(bits.length, _getBitSequenceCapacity(def.bitStoreType));
    final int byteCapacity =  (capacity / 8).ceil()
    final bitCapacity = byteCapacity * 8;
    final bitArray = BitArray(bitCapacity);

    encoder.compact(bits.length);

    for (var i = 0; i < bits.length; i++) {
      if(i > 1 || i < 0) { continue; } // Not sure if this is necessary
      bitArray[i] = i.isOdd;
    }
    encoder.bytes(bitArray.byteBuffer.asUint8List(0, byteCapacity).toList());
  }
```

packages/polkadart_scale_codec/lib/src/core/codec.dart
pastre: Added another more specific comment

packages/polkadart_scale_codec/lib/src/core/codec.dart
pastre: Nice error message!

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -453,18 +453,84 @@ class Codec {
...
+    final type = _types[bitStoreTypeIndex] as PrimitiveType;
+    final name = type.primitive.name.toLowerCase();
+    if (name.startsWith('u') == false) {
+      throw UnexpectedCaseException(
+          'Expected unsigned bit with `u` prefix, but got $name which can\'t be decoded');
+    }
+    if (name.length == 1 || int.parse(name.substring(1)) % 8 != 0) {
+      throw UnexpectedCaseException(
+          'Expected unsigned bit with valid bitSize, but got ${name.length == 1 ? '' : name.substring(1)} which can\'t be decoded.');
+    }
```
pastre, line 531: I think it's weird to have an empty string when we actually got `'u'`. `name` should provide context for the user on what's wrong
```suggestion
          'Expected unsigned bit with valid bitSize, but got $name which can\'t be decoded.');
    }
```

Issue #183, opened 2023-01-02
Re-code EraExtrinsic Decode/Encode (Generic way)


justkawal:
- [ ] EraExtrinsic Decode/Encode Generic Approach
- [ ] Remove the EraExtrinsic from PrimitiveTypes

Issue #178, opened 2022-12-26
Mortality Era Incorrect decoding


justkawal:
- Resolve incorrect missing decoding in Era : Mortality.

PR #184, opened 2023-01-02
Era resolution #178

justkawal:
## Any follow ups?
#180 

## Description
- A temporary fix to get `{"period": value, "phase": value}`.
- A issue is made to handle it in generic way: #183 
- Also improved `toHuman()` to handle `None` aka (`NoneOption`).

## Why?
#178 

## How?
- Done by manually extracting the era when making ChainDescription Types.

### DIFF
--- a/packages/polkadart_scale_codec/lib/src/core/codec.dart
+++ b/packages/polkadart_scale_codec/lib/src/core/codec.dart
@@ -545,6 +545,62 @@ class Codec {
     }
   }
 
+  ///
+  /// Decodes EraExtrinsic
+  Map<String, dynamic> _decodeExtrinsicEra(Source source) {
+    final optionData = source.bytes(1).toList();
+
+    if (optionData[0] == 0) {
+      return {'period': 0, 'phase': 0};
+    }
+
+    final bytes = decodeHex(
+      encodeHex(optionData).replaceAll(RegExp(r'0x'), '') +
+          encodeHex(source.bytes(1).toList()).replaceAll(RegExp(r'0x'), ''),
+    ).toList();
+
+    final encoded = bytes[0] + (bytes[1] << 8);
+
+    final period = 2 << (encoded % (1 << 4));
+
+    final phase = (encoded >> 4) * max(period >> 12, 1);
+
+    return {'period': period, 'phase': phase};
+  }
+
+  ///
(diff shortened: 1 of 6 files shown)


packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -576,6 +632,9 @@ class Codec {
         return source.boolean();
       case Primitive.Str:
         return source.str();
+      // TODO: Resolve issue: #183 - Extrinsic Era
```
leonardocustodio, line 635: Should we remove this?

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -646,6 +705,10 @@ class Codec {
       case Primitive.U256:
         encoder.u256(value);
         return;
+      // TODO: Resolve issue: #183 - Extrinsic Era
```
leonardocustodio, line 708: Should we remove this?

packages/polkadart_scale_codec/lib/src/core/types.dart
```diff
@@ -48,6 +48,10 @@ enum Primitive {
 
   /// Single Character
   Char,
+
+  /// Mocking Extrinsic Era as primitive for quick fix.
+  /// TODO: Issue: #183 : Remove this quick-fix once Extrinsic Era is implemented in Generic Way
```
leonardocustodio, line 53: Should we remove this?

leonardocustodio (approved):

Lohann (changes requested):
I'm missing unit tests for this new introduced code, please at least test the `Primitive.ExtrinsicEra`

packages/polkadart_scale_codec/lib/src/core/types.dart
```diff
@@ -48,6 +48,10 @@ enum Primitive {
...
+  ExtrinsicEra,
```
Lohann, line 54: I don't like this hack solution, big chances of we miss this out here, why we need a quick fix for this?

leonardocustodio: We kind of need this for testing ;)
Having tests to ensure it works and that when we refactor everything is correct should be good for us.

Issue #185, opened 2023-01-04
Change polkadart-scale-codec approach to encode/decode


justkawal:
Tasks:
- [x] Isolated encoding / decoding of types.
- [x] Unit Test-cases for all types.

PR #186, opened 2023-01-05
Justkawal/complete new approach

justkawal:
#Resolves Issue:
- #185 

## Description
- A complete new configurable re-design
- Easy to add functionality.
- Well documented

## Why?
- More clean approach
- Less boilerplate code
- Short, clear and concised
- Less dependency on other libraries

## How?
- By isolating the decode and encode functionality inside the basic type classes.
- Lesser intermediate cyclic conversion of types: `forth and back`.

## Testing Plan
- Unit Test cases.
- Large Unit Testing
- Command:  `dart test`


packages/polkadart_scale_codec/lib/src/core/type_registry.dart
```diff
@@ -0,0 +1,195 @@
...
+          if (match.length > 2) {
+            switch (match[1].toString().toLowerCase()) {
+              // vec array
+              case 'vec':
+              // option
+              case 'option':
+              // compact
+              case 'compact':
+              // BTreeMap
+              case 'BTreeMap':
```
Lohann, line 126: This case will never the be catch, because you do `.toLowerCase()` so is impossible to get `BTreeMap`.

Also I don't think we should do a lowercase here, because struct names are case sensitive, `BTreeMap` and `BTreemap` are considered different structs.

Lohann: Also this is bug, as such must have a unit tests covering this case, it should be catch by unit tests.

packages/polkadart_scale_codec/lib/src/core/type_registry.dart
```diff
@@ -0,0 +1,195 @@
...
+          break;
+        }
+        if (iterationSolve) {
+          continue;
+        }
+
+        // Complex type
+        if (value[-1] == '>') {
+          final match =
+              RegExp(r'^([^<]*)<(.+)>$').allMatches(value as String).toList();
```
Lohann, line 115: What that REGEX actually try to find? can you document it please? also I think this should be moved to a separated method and properly tested.

Lohann (changes requested):

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
```diff
@@ -0,0 +1,46 @@
+part of codec_types;
+
+///
+/// BoolCodec to encode/decode bool
+class BoolCodec extends Codec<bool> {
```
Lohann, line 5: Please create a test specifically for this struct

packages/polkadart_scale_codec/lib/src/core/codec_interface.dart
```diff
@@ -0,0 +1,13 @@
...
+///
+/// An interface for the codec
+mixin CodecInterface<T> {
+  ///
+  /// decodes the value form the source
+  dynamic decode();
+
+  ///
+  /// encodes the passed parameter and returns the encoded value
+  dynamic encode(T value);
```
Lohann, line 12: We must avoid using `dynamic` type, instead we must use strict types.

packages/substrate_metadata/test/dummy_test.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:test/test.dart';
+
+void main() {
+  group('dummy group', () {
+    test('dummy test', () {
+      expect(1, 1);
+    });
+  });
```
Lohann, line 8: ...

justkawal: because when the workflow will run dart test in every directory then it will crash the Github Actions

packages/polkadart_scale_codec/lib/src/core/type_registry.dart
justkawal: Resolved ✅

packages/polkadart_scale_codec/lib/src/core/type_registry.dart
justkawal: Done ✅
Added test-cases
Explained some few things on comment when accessing match[2] when groupCount is just 2.

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
justkawal: Test Cases already present:
- With single direct type call.
- With Custom User defined structure of json.

packages/polkadart_scale_codec/lib/src/core/codec_interface.dart
justkawal: Done removed usage of dynamic and change it to generic `<T>`

packages/polkadart_scale_codec/test/type_registry_test.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Test TypeRegistry', () {
+    test('When registry is created then it should have length 1', () {
```
pastre, line 6: Why should it have lenght 1?

packages/polkadart_scale_codec/test/type_registry_test.dart
```diff
@@ -0,0 +1,33 @@
...
+    test('When registry is created then it should have the defined keys.', () {
+      final keys = ['bool'];
+
+      final Registry registry = TypeRegistry.createRegistry();
+
+      final evaluatedKeys = registry.keys.toList();
+      expect(evaluatedKeys, equals(keys));
+    });
+
+    test('When map of codecs is added to registry then it should have length 2',
```
pastre, line 20: why?

packages/polkadart_scale_codec/test/utils/extensions_test.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  /// extensions test cases
+  group('Test String extension', () {
+    test('When getter first is called on \'Hello\' then \'H\' is returned', () {
```
pastre, line 7:
```suggestion
    test('first should return first character', () {
```

packages/polkadart_scale_codec/test/utils/extensions_test.dart
```diff
@@ -0,0 +1,29 @@
...
+      final firstChar = 'Hello'.first;
+      expect(firstChar, equals('H'));
+    });
+
+    test('When getter last is called on \'Hello\' then \'o\' is returned', () {
```
pastre, line 12:
```suggestion
    test('last is should return last character', () {
```

packages/polkadart_scale_codec/test/utils/extensions_test.dart
```diff
@@ -0,0 +1,29 @@
...
+      final lastChar = 'Hello'.last;
+      expect(lastChar, equals('o'));
+    });
+
+    test(
+        'When getter first is called on \'\' then UnexpectedCaseException is thrown',
```
pastre, line 18:
```suggestion
        'given an empty string when getting first character it should throw',
```

packages/polkadart_scale_codec/test/utils/extensions_test.dart
```diff
@@ -0,0 +1,29 @@
...
+        () {
+      expect(() => ''.first, throwsA(isA<EmptyStringException>()));
+    });
+
+    test(
+        'When getter last is called on \'\' then UnexpectedCaseException is thrown',
```
pastre, line 24:
```suggestion
        'given an empty string when getting last character it should throw',
```

packages/polkadart_scale_codec/test/utils/hex_test.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Hex Encode Test: ', () {
```
pastre, line 5: When unit testing this, you shouldnt test values. The way you did it, you're using (and trusting) dart's `HexCodec`, so you only need to test if you're calling `hex.encode()`. Testing values is overall a bad idea in this case bc we're testing functionality that we don't maintain instead of testing the integration between components

packages/polkadart_scale_codec/test/utils/hex_test.dart
```diff
@@ -0,0 +1,42 @@
...
+      expect(decodedValue, equals([1]));
+    });
+
+    test('When 0x00 is decoded then hex value is [0]', () {
+      final decodedValue = decodeHex('00');
+      expect(decodedValue, equals([0]));
+    });
+  });
+
+  group('Hex Invalid Input Test: ', () {
```
pastre, line 29: Test for `assertionCheck`?

packages/substrate_metadata/lib/substrate_metadata.dart
```diff
@@ -0,0 +1 @@
+class Aweosme {}
```
pastre, line 1: ?

packages/polkadart_scale_codec/test/codec/bool_test.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  /// bool codec test cases
+  group('Test bool Decode', () {
```
pastre, line 6: This is not testing bool decode. It's testing that:

1. TypeRegistry.createRegistry() adds bool to its registry by default (which means bool is in the default array)
2. Codec initializes with registry
3. Codec.createTypeCodec converts the type
4. Codec.createTypeCodec finds a matching type name
5. Codec.createTypeCodec properly creates metadata and sets typeString of the new codec
6. BoolCodec decodes properly

In your test, you're only asserting for number 6, which is the only thing actually being covered. This test  will accuse more coverage then it actually covers
This test will fail if any change impacts numbers 1-5, causing it to be flaky

Please test `BoolCodec` directly

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
```diff
@@ -0,0 +1,63 @@
...
+  final Source? source;
+
+  ///
+  /// constructor
+  BoolCodec({Registry? registry, this.source})
+      : super(registry: registry ?? Registry());
+
```
pastre, line 12: Why is this depending on the registry? Also, it shouldn't be optional

pastre (changes requested):

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
```diff
@@ -0,0 +1,62 @@
+part of codec_types;
+
+///
+/// BoolCodec to encode/decode bool
+class BoolCodec extends Codec<bool> {
+  final Source? source;
```
Lohann, line 6: I really don't like this `Source` being here, I think this must be provided as parameter, not stored in the Codec.
I don't see any advantage on storing the source together in the Codec.

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
```diff
@@ -0,0 +1,62 @@
...
+
+    if (bytes[0] == 1) {
+      return true;
+    } else if (bytes[0] == 0) {
+      return false;
+    }
+
+    throw UnexpectedCaseException(
+        'Expected 0 or 1 but found byte: ${bytes[0]}');
+  }
```
Lohann, line 40: Simply call `source.byte()` method here:
```dart
  bool decode() {
    final byte = (source ?? data).byte();
    return byte > 0;
  }
```

packages/polkadart_scale_codec/lib/src/codec_types/bool_codec.dart
```diff
@@ -0,0 +1,62 @@
...
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<bool>().createTypeCodec('Bool');
+  /// final boolValue = codec.encode(false);
+  /// print(boolValue); // 00
+  /// ```
+  @override
+  String encode(bool value) {
+    return value ? '01' : '00';
+  }
```
Lohann, line 61: Nop.. we should not use hex strings for encoding, this is very inefficient.

packages/polkadart_scale_codec/lib/src/core/codec_interface.dart
```diff
@@ -0,0 +1,13 @@
...
+/// An interface for the codec
+mixin CodecInterface<T> {
+  ///
+  /// decodes the value form the source
+  T decode();
+
+  ///
+  /// encodes the passed parameter and returns the encoded value
+  String encode(T value);
+}
```
Lohann, line 13: Where we use this? why not use Codec type?

Lohann (changes requested):

packages/polkadart_scale_codec/test/codec/u8_test.dart
```diff
@@ -0,0 +1,152 @@
...
+      final u8Value = codec.decode();
+      expect(u8Value, equals(0));
+    });
+
+    test('When highest value 0xff is decoded then it returns 255', () {
+      final codec =
+          Codec(registry: registry).createTypeCodec('u8', input: Input('0xff'));
+      final u8Value = codec.decode();
+      expect(u8Value, equals(255));
+    });
```
Lohann, line 19: I don't think the must provide the `input` when creating the codec, the codec and the input must be decoupled.
```dart
    test('When highest value 0xff is decoded then it returns 255', () {
      final codec = Codec(registry: registry).createTypeCodec('u8');
      final u8Value = codec.decode(input: Input('0xff'));
      expect(u8Value, equals(255));
    });
```

The input should not be stored in the codec

packages/polkadart_scale_codec/lib/src/core/type_registry.dart
```diff
@@ -0,0 +1,180 @@
+part of core;
+
+///
+/// Converting above php code to dart code
+class TypeRegistry {
+  ///
+  /// Basic Codec names
+  static const _defaultCodecTypes = <String>['Bool', 'U8'];
```
Lohann, line 8: ? where we this, and why those are the defaultCodecTypes? what basic codec names means?

packages/polkadart_scale_codec/lib/src/core/type_registry.dart
```diff
@@ -0,0 +1,172 @@
...
+    if (codec != null) {
+      return codec;
+    }
+
+    throw UnexpectedCodecException('Unable to find codec for $codecName');
+  }
+
+  ///
+  /// Parses customJson and adds it to registry
+  static void addCustomCodec(
```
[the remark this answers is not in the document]
Lohann: I agree, this function is really big and complex, and requires more testing.

packages/polkadart_scale_codec/lib/src/core/codec_interface.dart
```diff
@@ -0,0 +1,13 @@
+part of core;
+
+///
+/// An interface for the codec
+mixin CodecInterface<T> {
```
Lohann, line 5: I still don't understand where and why we need this interface

packages/substrate_metadata/test/dummy_test.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:test/test.dart';
+
+void main() {
+  group('dummy test', () {
+    test('dummy test', () {
+      expect(1, 1);
+    });
+  });
+}
```
Lohann, line 9: Why we need this dummy test? can't we simply remove the this folder completly?

packages/substrate_metadata/lib/substrate_metadata.dart
Lohann: ?

Lohann (changes requested):

packages/polkadart_scale_codec/lib/primitives/h256.dart
```diff
@@ -0,0 +1,20 @@
+part of primitives;
+
+class H256Codec with Codec<Uint8List> {
+  const H256Codec._();
+
+  static final H256Codec instance = H256Codec._();
```
Lohann, line 6: Why not simply:
```dart
final H256Codec = ArrayCodec(U8Codec.instance, 32);
```
lets avoid create unnecessary primitive types.

packages/polkadart_scale_codec/lib/primitives/option.dart
```diff
@@ -1,13 +1,63 @@
-// ignore_for_file: annotate_overrides
+// ignore_for_file: library_private_types_in_public_api
 
-part of polkadart_scale_codec_core;
+part of primitives;
 
+class OptionCodec with Codec<OptionType> {
```
Lohann, line 5: This seems a bit overengineered, why not simply this:
```dart
class OptionCodec<E> with Codec<E?> {
  final Codec<E> subtypeCodec;

  const OptionCodec(this.subtypeCodec);

  @override
  void encodeTo(E? maybeValue, Output dest) {
    if (maybeValue == null) {
      dest.pushByte(0);
    } else {
      dest.pushByte(1);
      subtypeCodec.encodeTo(maybeValue, dest);
    }
  }

  @override
  E? decode(Input input) {
    if (input.read() == 0) {
      return null;
    }
    return subtypeCodec.decode(input);
  }

  @override
  int sizeHint(E? maybeValue) {
    if (maybeValue == null) {
      return 1;
    }
    return 1 + subtypeCodec.sizeHint(maybeValue);
  }
}
```

packages/polkadart_scale_codec/lib/primitives/result.dart
```diff
@@ -0,0 +1,43 @@
+part of primitives;
+
+///
+/// Result to encode/decode map of values
+class ResultCodec<T> with Codec<Map<String, T>> {
```
Lohann, line 5: Do we really need a Result codec? can't we simply use an variant?
```dart
VariantCodec({
  'Ok': ObjOk.instance,
  'Err': ObjErr.insntace,
});
```

packages/polkadart_scale_codec/lib/primitives/struct.dart
```diff
@@ -0,0 +1,42 @@
+part of primitives;
+
+///
+/// Struct to encode/decode map of values
+class StructCodec with Codec<LinkedHashMap<String, dynamic>> {
```
Lohann, line 5: Lets use the same name as `scale-info`, instead of `Struct` lets call it `CompositeCodec`:
https://github.com/paritytech/scale-info/blob/v2.3.1/src/ty/mod.rs#L276-L278

packages/polkadart_scale_codec/lib/primitives/bytes.dart
```diff
@@ -0,0 +1,19 @@
+part of primitives;
+
+class BytesCodec with Codec<Uint8List> {
+  const BytesCodec._();
+
+  static final BytesCodec instance = BytesCodec._();
```
Lohann, line 6: We don't need this type too, we can simply use:
```dart
SequenceCodec(U8Codec.instance)
```

Is a bad practice having two ways for doing the same thing.

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -0,0 +1,58 @@
+part of primitives;
+
+class SimpleEnumCodec<A> with Codec<A> {
```
Lohann, line 3: We can simplify this code, instead of two codec `SimpleEnumCodec ` and `ComplexEnumCodec` which increase complexity, lets have just one `VariantCodec`, which accepts null as codec:
```dart
VariantCodec({
  'A': null,
  'B': U8Codec.instance,
  'C': PointCodec.instance,
})
```
We can have a specialized constructor for simple enum:
```dart
VariantCodec.fromValues(['VariantA', 'VariantB', 'VariantC'])
// same as
VariantCodec({
  'VariantA': null,
  'VariantB': null,
  'VariantC': null,
})
```

Lohann (changes requested):

packages/polkadart_scale_codec/lib/primitives/option.dart
Lohann: Btw, for dynamic, an `Option` is simply an enum:
```dart
VariantCodec({
  'None': null,
  'Some': ValueCodec.instance,
})
```

packages/polkadart_scale_codec/lib/primitives/str.dart
```diff
@@ -0,0 +1,31 @@
...
+    return String.fromCharCodes(elements);
+  }
+
+  @override
+  int sizeHint(String value) {
+    var size = CompactCodec.instance.sizeHint(value.length);
+    size += value.codeUnits.length;
+    return size;
+  }
+}
```
Lohann, line 31: Test are missing for this one, is also wrong, it MUST convert the string to utf-8:

```dart
import '../scale_codec.dart' show Codec, Input, Output;
import './compact.dart' show CompactCodec;
import './u8.dart' show U8Codec;
import 'dart:convert' show utf8;

class StrCodec with Codec<String> {
  const StrCodec._();

  static const StrCodec codec = StrCodec._();

  @override
  void encodeTo(String str, Output dest) {
    final utf8str = utf8.encode(str);
    CompactCodec.codec.encodeTo(utf8str.length, dest);
    for (final charCode in utf8str) {
      U8Codec.codec.encodeTo(charCode, dest);
    }
  }

  @override
  String decode(Input input) {
    final size = CompactCodec.codec.decode(input).toInt();
    final bytes = input.readBytes(size).toList();
    return utf8.decode(bytes);
  }

  @override
  int sizeHint(String str) {
    var size = CompactCodec.codec.sizeHint(str.length);
    size += utf8.encode(str).length;
    return size;
  }
}
```

packages/polkadart_scale_codec/lib/primitives/null_codec.dart
```diff
@@ -0,0 +1,14 @@
...
+  const NullCodec._();
+
+  static const NullCodec instance = NullCodec._();
+
+  @override
+  void decode(Input input) {}
+
+  @override
+  void encodeTo(dynamic value, Output output) {}
+}
```
Lohann, line 14: Instead of dynamic, use void, so the compiler knows the element have size equals to zero:
- https://medium.com/flutter-community/the-curious-case-of-void-in-dart-f0535705e529
```dart
import '../scale_codec.dart' show Codec, Input, Output;

class NullCodec with Codec<void> {
  const NullCodec._();

  static const EmptyCodec codec = EmptyCodec._();

  @override
  void encodeTo(void element, Output output) {
    return;
  }

  @override
  void decode(Input input) {}

  @override
  int sizeHint(void element) {
    return 0;
  }
}
```

packages/polkadart_scale_codec/lib/primitives/option.dart
```diff
@@ -1,13 +1,51 @@
...
+/// OptionType
 ///
-/// Making Option private and hence only `Some()` and `None` will be exposed for usage
-class _Option<T> extends Equatable {
+/// OptionType is a generic type that can be either Some(T) or None.
+class OptionType<T> extends Equatable {
   final String kind;
   final T? value;
-  const _Option(this.kind, [this.value]);
+  const OptionType(this.kind, [this.value]);
```
Lohann, line 48: There's a bug here if `OptionType<T?>`, to fix that use this:
```

File: packages/substrate_metadata/test/dummy_test.dart (hunk 25 above)
Comment (Lohann) [reply packages/substrate_metadata/test/dummy_test.dart:9]: Please remove the dummy test

File: packages/polkadart_scale_codec/lib/primitives/compact.dart [hunk 35]
```diff
@@ -0,0 +1,165 @@
...
+    } else if (value < 16384) {
+      output
+        ..pushByte((value.toInt() & 0x3F) << 2 | 1)
+        ..pushByte((value.toInt() >> 6).toUnsigned(8));
+    } else if (value < 1073741824) {
+      output
+        ..pushByte((value.toInt() & 0x3F) << 2 | 2)
+        ..pushByte((value.toInt() >> 6).toUnsigned(8))
+        ..pushByte((value.toInt() >> 14).toUnsigned(8))
+        ..pushByte((value.toInt() >> 22).toUnsigned(8));
```
Lohann, line 103: The value is already of type int here
```dart
      output.pushByte(value << 2);
    } else if (value < 16384) {
      output
        ..pushByte((value & 0x3F) << 2 | 1)
        ..pushByte((value >> 6).toUnsigned(8));
    } else if (value < 1073741824) {
      output
        ..pushByte((value & 0x3F) << 2 | 2)
        ..pushByte((value >> 6).toUnsigned(8))
        ..pushByte((value >> 14).toUnsigned(8))
        ..pushByte((value >> 22).toUnsigned(8));
```

packages/polkadart_scale_codec/lib/primitives/compact.dart
```diff
@@ -0,0 +1,165 @@
...
+    } else {
+      assertion(value.bitLength >= 30,
+          'Previously checked anyting less than 2^30; qed');
+      final bytesNeeded = (value.bitLength + 7) >> 3;
+      output.pushByte(((bytesNeeded - 4) << 2).toUnsigned(8) | 3);
+      for (var i = 0; i < bytesNeeded; i++) {
+        output.pushByte(value.toUnsigned(8).toInt());
```
Lohann, line 110:
```suggestion
        output.pushByte(value.toUnsigned(8));
```

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -0,0 +1,76 @@
+part of primitives;
+
+class SimpleEnumCodec<A> with Codec<A?> {
+  final List<A?> list;
```
Lohann, line 4: This should be an Map, because even simple enums can have sparse indexes, ex:

```rust
enum SomeEnum {
  #[index(0)]
  Value1,
  #[index(10)]
  Value2,
  #[index(30)]
  Value3,
}
```

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -0,0 +1,76 @@
...
+    if (list[index] == null) {
+      throw EnumException('Value at index: $index is null, $index not usable.');
+    }
+
+    return list[index]!;
+  }
+}
+
+class ComplexEnumCodec<V> with Codec<MapEntry<String, V?>> {
+  final Map<String, Codec<V?>?> map;
```
Lohann, line 45: Same case here, where do you store the enum index?

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -0,0 +1,76 @@
...
+    if (index == -1) {
+      throw EnumException(
+          'Invalid enum value: $value. Can only accept: ${map.keys.toList()}');
+    }
+
+    output.pushByte(index);
+
+    final codec = map[value.key];
+
+    codec?.encodeTo(value.value, output);
```
Lohann, line 61: Unnecessary line break.
```suggestion
    output.pushByte(index);
    final codec = map[value.key];
    codec?.encodeTo(value.value, output);
```

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -0,0 +1,76 @@
...
+  MapEntry<String, V?> decode(Input input) {
+    final index = input.read();
+    if (index >= map.length) {
+      throw EnumException('Invalid enum index: $index.');
+    }
+    final key = map.keys.elementAt(index);
+
+    final codec = map[key];
+
+    return MapEntry(key, codec?.decode(input));
```
Lohann, line 74:
```suggestion
    final key = map.keys.elementAt(index);
    final codec = map[key];
    return MapEntry(key, codec?.decode(input));
```

packages/polkadart_scale_codec/test/direct_tests/array_test.dart
```diff
@@ -0,0 +1,107 @@
...
+      final output = HexOutput();
+      ArrayCodec(
+        ArrayCodec(U8Codec.instance, 2),
+        2,
+      ).encodeTo([
+        [1, 2],
+        [3, 4]
+      ], output);
+      expect(output.toString(), '0x01020304');
+    });
```
Lohann, line 23: Nice ❤️

packages/polkadart_scale_codec/test/direct_tests/i16_test.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('I16 Decode Test:', () {
```
Lohann, line 5: We also need tests for `OutOfBounds` errors

packages/polkadart_scale_codec/test/direct_tests/i256_test.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('I256 Decode Test:', () {
```
Lohann, line 5: We also need tests for `OutOfBounds` errors

packages/polkadart_scale_codec/test/direct_tests/i128_test.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:polkadart_scale_codec/polkadart_scale_codec.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('I128 Decode Test:', () {
```
Lohann, line 5: We also need tests for `OutOfBounds` errors

packages/polkadart_scale_codec/test/direct_tests/option_test.dart
```diff
@@ -0,0 +1,106 @@
...
+      expect(result, Option.some(Option.some(true)));
+    });
+
+    test(
+        'Given a 0x010100 it should be decoded to Option.some(Option.some(false))',
+        () {
+      final input = HexInput('0x010100');
+      final result = OptionCodec(OptionCodec(BoolCodec.instance)).decode(input);
+      expect(result, Option.some(Option.some(false)));
+    });
```
Lohann, line 104: Nice

packages/polkadart_scale_codec/test/direct_tests/simple_enum_test.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  group('SimpleEnumCodec Encode', () {
+    test('should encode and decode', () {
+      final output = HexOutput();
+      final codec = SimpleEnumCodec(['a', 'b', 'c']);
+      codec.encodeTo('b', output);
+      expect(output.toString(), '0x01');
+    });
```
Lohann, line 11: We need tests for sparse indexes, I think in this case we can do this:
```dart
SimpleEnumCodec.sparse({
  'a': 0,
  'b': 10,
  'c': 20,
});
```

Lohann (changes requested):

packages/polkadart_scale_codec/test/direct_tests/simple_enum_test.dart
Lohann: Or the inverse (so we guarantee the index is unique):
```dart
SimpleEnumCodec.sparse({
  0: 'a',
  10: 'b'
  20: 'c',
});
```

packages/polkadart_scale_codec/lib/io/byte_input.dart
```diff
@@ -0,0 +1,27 @@
...
+    _buffer = buffer;
+  }
+
+  ByteInput.fromList(List<int> buffer) {
+    _buffer = Uint8List.fromList(buffer);
+  }
+
+  @override
+  int read() {
+    return _buffer[offset++];
```
Lohann, line 15: Wondering where the offset is stored

packages/polkadart_scale_codec/lib/io/hex_input.dart
```diff
@@ -0,0 +1,19 @@
+part of io;
+
+class HexInput with Input {
+  HexInput(String hex) {
+    _buffer = decodeHex(hex);
+  }
+
+  @override
+  int read() {
+    return _buffer[offset++];
```
Lohann, line 10: Same case here, where this offset is stored?

packages/polkadart_scale_codec/lib/primitives/option.dart
```diff
@@ -0,0 +1,60 @@
...
+        throw OptionException('Invalid Option Byte: $b');
+    }
+  }
+}
+
+const None = Option.none();
+
+class Option extends Equatable {
+  final dynamic value;
+  final bool _isSome; // Workaround for Option<int?>.some(null)
```
Lohann, line 40:
```suggestion
class Option<T> extends Equatable {
  final T value;
  final bool _isSome; // Workaround for Option<int?>.some(null)
```

PR #223, opened 2023-01-06
Re-design - u8 (New Approach)

justkawal:
## Any follow ups?
#186 

## Description
- Adds u8 encode and decode functionality.
- Large quantity of test-cases
- Test cases to test u8

## Why?
- Solves #190 

## Testing Plan 
`dart test`

### DIFF
--- a/packages/polkadart_scale_codec/lib/src/codec_types/u8.dart
+++ b/packages/polkadart_scale_codec/lib/src/codec_types/u8.dart
@@ -1,0 +1,69 @@
+part of codec_types;
+
+///
+/// U8 to encode/decode unsigned 8 bit integer
+class U8 extends Codec<int> {
+  ///
+  /// constructor
+  U8({Registry? registry}) : super(registry: registry ?? Registry());
+
+  ///
+  /// Decode a unsigned 8 bit integer from the source
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8', data: Source('0x01'));
+  /// final value = codec.decode();
+  /// print(value); // 1
+  /// ```
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8', data: Source('0x00'));
+  /// final value = codec.decode();
+  /// print(value); // 0
+  /// ```
+  @override
+  int decode() {
+    return bytesToLittleEndianInt(data.bytes(1).toList());
+  }
+
+  ///
+  /// Encodes a unsigned 8 bit integer
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8');
+  /// final value = codec.encode(1);
+  /// print(value); // 01
+  /// ```
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8');
+  /// final value = codec.encode(0);
+  /// print(value); // 00
+  /// ```
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8');
+  /// final value = codec.encode(255);
+  /// print(value); // ff
+  /// ```
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8');
+  /// final value = codec.encode(256);
+  /// print(value); // 00
+  /// ```
+  @override
+  String encode(int value) {
+    if (value >= 0 && value <= 255) {
+      return encodeHex(littleEndianIntToBytes(value, 1));
+    }
+    throw UnexpectedCaseException(
+        'Expected value between 0 and 255, but found: $value');
+  }
+}
--- a/packages/polkadart_scale_codec/lib/src/utils/functions.dart
+++ b/packages/polkadart_scale_codec/lib/src/utils/functions.dart
@@ -8,3 +8,46 @@ void assertionCheck(bool value, [String? msg]) {
     throw AssertionException(msg ?? 'Assertion Error occured.');
   }
 }
+
+/// Converts a [List<int>] to [int]
+///
+/// Example:
+/// ```dart
+/// final bytes = [1];
+/// final val = bytesToLittleEndianInt(bytes);
+/// print(val); // 1
+/// ```
+///
+/// Example:
+/// ```dart
+/// final bytes = [1, 2];
+/// final val = bytesToLittleEndianInt(bytes);
+/// print(val); // 258
+/// ```
+///
+/// Example:
+/// ```dart
+/// final bytes = [1, 2, 3, 4];
+/// final val = bytesToLittleEndianInt(bytes);
+/// print(val); // 16909060
+/// ```
+int bytesToLittleEndianInt(List<int> bytes) {
+  switch (bytes.length) {
+    case 1:
+      return bytes[0];
+    case 2:
+      return bytes[0] | bytes[1] << 8;
+    case 4:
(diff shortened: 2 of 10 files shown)


packages/polkadart_scale_codec/lib/src/codec_types/u8.dart
```diff
@@ -0,0 +1,57 @@
...
+  ///
+  /// constructor
+  U8({this.source}) : super(registry: Registry());
+
+  ///
+  /// Decode a unsigned 8 bit integer from the source
+  ///
+  /// Example:
+  /// ```dart
+  /// final codec = Codec<int>().createTypeCodec('U8', data: Source('0x00'));
```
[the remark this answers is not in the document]
gabrielokura: Or examples to more complex structures: 

```dart
final codec = CodecTuple(CodecI8, CodecU16);
encoded = codec.encode(data1, data2);
decoded = codec.decode(encoded);
```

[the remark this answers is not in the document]
justkawal: You could also use it in manner:

```dart
final codec = U8(source: Source('0xhex_value'));
final decodedValue = codec.decode();

// for encoding.
final encodedValue = U8().encode(value to encode);
```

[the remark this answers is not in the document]
justkawal: I'll stick with the current implementation....

Lohann (approved):

packages/polkadart_scale_codec/lib/src/utils/functions.dart
```diff
@@ -8,3 +8,42 @@ void assertionCheck(bool value, [String? msg]) {
...
+    case 8:
+      {
+        var byte = bytes[0];
+        for (var index = 1; index < bytes.length; index++) {
+          byte |= bytes[index] << (8 * index);
+        }
+        return byte;
+      }
+  }
+  return 0;
```
Lohann, line 48: It doesn't work for 8 bytes (64bit), DART integer type can only store up to `53bit` numbers when running in `web`:
https://dart.dev/guides/language/numbers#precision

Please use BigInt or limit the number of bits for `int` type

packages/polkadart_scale_codec/test/utils/bytes_to_little_endian_int_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        () {
+      final intValue = bytesToLittleEndianInt([0, 0, 1, 0, 0, 0, 0]);
+      expect(intValue, equals(0));
+    });
+
+    test(
+        'When bytes of length greator than 8 is evaluated irrespective of bytes then it returns 0',
+        () {
+      final intValue = bytesToLittleEndianInt([0, 0, 1, 0, 0, 0, 0, 0, 0]);
+      expect(intValue, equals(0));
```
Lohann, line 135: I don't think it should return `0`, it must return an exception or `null` to make explicitly that is not possible to represent this value

Lohann (changes requested):

packages/polkadart_scale_codec/lib/src/utils/functions.dart
Lohann: Also doesn't limit for power of `2`, the byte.length can be any number, so you can use this for decode `compact` number too.

packages/polkadart_scale_codec/test/utils/bytes_to_little_endian_int_test.dart
justkawal: Fixed

packages/polkadart_scale_codec/lib/src/utils/functions.dart
justkawal: Done

packages/polkadart_scale_codec/lib/src/core/codec.dart
```diff
@@ -175,7 +175,21 @@ class Codec<T> implements CodecInterface<T> {
...
+  ///
+  /// `throws [EOSException] if the input.remainingLength > 0`
+  ///
+  /// Please note that this method is useful only when the input is decoded from a Input.
+  ///
+  /// If the input is being encoded, then this method is not useful.
+  void assertEOS() {
+    if (input.remainingLength > 0) {
+      throw EOSException();
+    }
```
Lohann, line 193: I don't understand this error, shouldn't throw this when `remainingLength == 0`??

packages/polkadart_scale_codec/lib/src/core/input.dart
```diff
@@ -0,0 +1,99 @@
+part of core;
+
+class Input {
```
Lohann, line 3: I think the input should be a `mixin` not a class, so we can have different sources in the future `network, websockets, etc`

justkawal: Alright...

packages/polkadart_scale_codec/lib/src/core/input.dart
```diff
@@ -0,0 +1,99 @@
...
+  /// Get the bytes from the current index to the length
+  ///
+  /// Example:
+  /// ```dart
+  /// final input = Input('0x010203');
+  /// print(input.bytes(3)); // [1, 2, 3]
+  /// ```
+  Uint8List bytes(int length) {
+    final beg = currentIndex;
+    final end = currentIndex += length;
```
Lohann, line 74: are you incrementing the currentIndex at the same time while we assign the value? I think it will be more clear to split it in two lines:
```dart
final beg = currentIndex;
currentIndex += length;
final end = currentIndex;
```

packages/polkadart_scale_codec/lib/src/core/input.dart
```diff
@@ -0,0 +1,99 @@
...
+  /// Input(buffer_to_decode)
+  ///
+  /// Example:
+  /// ```dart
+  /// final input = Input('0x010203');
+  /// print(input.buffer); // [1, 2, 3]
+  /// ```
+  Input(String buffer) {
+    this.buffer = decodeHex(buffer);
+  }
```
Lohann, line 22: I think this one should be called `Input.fromHex(String buffer)`

packages/polkadart_scale_codec/lib/src/core/codec.dart
Lohann: Please create a test case of this error

packages/polkadart_scale_codec/lib/src/core/input.dart
justkawal: Fixing in #245

packages/polkadart_scale_codec/lib/src/core/input.dart
justkawal: Fixing in #245

packages/polkadart_scale_codec/lib/src/core/input.dart
justkawal: Corrected

packages/polkadart_scale_codec/lib/src/utils/functions.dart
```diff
@@ -8,3 +8,40 @@ void assertionCheck(bool value, [String? msg]) {
...
+  if (bytes.isEmpty) {
+    throw AssertionException('Bytes cannot be empty');
+  }
+  if (bytes.length < 9) {
+    var byte = bytes[0];
+    for (var index = 1; index < bytes.length; index++) {
+      byte |= bytes[index] << (8 * index);
+    }
+    return byte;
+  }
```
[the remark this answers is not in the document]
justkawal: Not using anymore....

Lohann (approved):

PR #258, opened 2023-02-28
Legacy parser for V9-V13

justkawal:
## Any follow ups?
#257 

## Description
- Adds Set implementation
- Resolves Indexed Enum in the Registry
- Adds implementation for Legacy Parser to create `event` and `extrinsics` lookup.

## Testing
- `dart test`

### DIFF
--- a/packages/polkadart_scale_codec/lib/core/registry.dart
+++ b/packages/polkadart_scale_codec/lib/core/registry.dart
@@ -327,36 +327,50 @@ class Registry {
     late int bitLength;
     if (value.containsKey('_bitLength')) {
       bitLength = value['_bitLength']!;
-      value.remove('_bitLength');
     } else {
       bitLength = 8;
     }
     assertion(bitLength % 8 == 0, 'Set bitLength should be multiple of 8.');
     assertion(bitLength > 0, 'Set bitLength should be greater than 0');
 
-    final codec = SetCodec(bitLength, value.keys.toList());
+    final values = value.entries
+        .where((element) => element.key != '_bitLength')
+        .map((e) => e.key)
+        .toList();
+
+    final codec = SetCodec(bitLength, values);
     addCodec(key, codec);
     return codec;
   }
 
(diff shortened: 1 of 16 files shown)


packages/polkadart_scale_codec/lib/core/registry.dart
```diff
@@ -117,6 +117,8 @@ class Registry {
               return _parseResult(customJson, key, value);
             case 'BitVec':
               return _parseBitSequence(customJson, key, value);
+            case 'DoNotConstruct':
+              return NullCodec.instance;
```
Lohann, line 121: ? this is for internal use only?

packages/polkadart_scale_codec/lib/primitives/enum.dart
```diff
@@ -1,11 +1,12 @@
 part of primitives;
 
 class SimpleEnumCodec<A> with Codec<A> {
-  final List<A> list;
+  final List<A?> list;
```
Lohann, line 4: Why not using a map instead an array?

justkawal:
```
 final output = HexOutput();
 final codec = SimpleEnumCodec(['a', 'b', 'c']);
 codec.encodeTo('b', output);
 expect(output.toString(), '0x01');
```

It is useful to automatically pick the index

Also don't have to do: ` codec.encodeTo(MapEntry('b', null), output);`

Can avoid writing null again and again.

It is changed to A? in the PR: #186 .

packages/polkadart_scale_codec/lib/core/registry.dart
justkawal: Yes, We will not export it and hide the type in the exporter file.

packages/polkadart_scale_codec/lib/primitives/enum.dart
justkawal: Resolved by forwarding changes to child PRs.

PR #285, opened 2023-03-21
Stack-overflow detection new Implementation

justkawal:
## Any follow ups?
#276 

## Description
- Adds Fix for #275 
- Performance improvement.

## Testing Plan 
`dart test`


packages/polkadart_scale_codec/lib/core/registry.dart
```diff
@@ -47,68 +58,51 @@ class Registry {
...
-        continue;
-      }
-
-      final codec = _parseCodec(customJson, key, value);
-      if (getCodec(key) == null) {
-        addCodec(key, codec);
-      }
+  void _callProxyLoaders() {
+    for (final proxy in _proxyLoaders.values) {
+      proxy();
```
Lohann, line 99:
```suggestion
    for (final loader in _proxyLoaders.values) {
      loader();
```

Lohann (changes requested):

packages/polkadart_scale_codec/lib/core/registry.dart
justkawal: Resolved

packages/polkadart_scale_codec/lib/primitives/proxy.dart
```diff
@@ -0,0 +1,42 @@
...
+///    'Orange': 'OrangeStruct',
+///  },
+/// }
+/// 
+/// 
+/// In above example, 'OrangeStruct' is calling 'JuiceEnumComplex' and 'JuiceEnumComplex' is calling 'OrangeStruct'.
+class ProxyCodec with Codec<dynamic> {
+  late final Codec codec;
+  ProxyCodec();
+
```
Lohann, line 32: Btw I also think is a good idea put here a `setter` and validate if the codec doesn't point to another Proxy:
```dart
void set codec(Codec codec) {
    if (codec is ProxyCodec) {
        throw Exception('Nested ProxyCodecs is not allowed');
    }
    _codec = codec;
}
```

Lohann (approved):

Lohann (approved):

PR #312, opened 2023-07-09
Fix Dart Generator JsonType cache

Lohann:
This PR fixes in the generator cache

# Bug
Due some `cacheKey` collisions, the Dart Generator was producing the wrong output for some chains, now we are using the substrate-metadata type id as cache key, which also improves the performance.

### DIFF
--- a/packages/polkadart_cli/lib/src/typegen/types/core.dart
+++ b/packages/polkadart_cli/lib/src/typegen/types/core.dart
@@ -5,41 +5,6 @@ typedef BasePath = String;
 abstract class TypeDescriptor {
   const TypeDescriptor();
 
-  static int _idSequence = 1;
-  static Map<TypeDescriptor, int> generatorToId = {};
-  static Map<String, TypeReference> jsonTypeCache = {};
-
-  static String _cachedKey(BasePath from, Set<TypeDescriptor> visited) {
-    bool created = false;
-    final ids = <int>[];
-    for (final generator in visited) {
-      int? id = generatorToId[generator];
-      if (id == null) {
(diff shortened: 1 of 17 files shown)


packages/polkadart_cli/lib/src/typegen/types/core.dart
```diff
@@ -5,41 +5,6 @@ typedef BasePath = String;
...
-    for (final generator in visited) {
-      int? id = generatorToId[generator];
-      if (id == null) {
-        id = _idSequence++;
-        generatorToId[generator] = id;
-        created = true;
-      }
-      ids.add(id);
-    }
-    ids.sort();
```
Lohann, line 24: The bug was here, the `ids.sort();` was causing the collisions, this logic is removed because the Generator now have the `id()` attribute, which is already global unique.

packages/polkadart_cli/lib/src/typegen/types/core.dart
```diff
@@ -172,3 +132,42 @@ class Field {
...
+  final Set<TypeDescriptor> _visited = {};
+  bool _isCircular = false;
+
+  TypeBuilderContext({required this.from});
+
+  TypeReference jsonTypeFrom(TypeDescriptor descriptor) {
+    if (_isCircular) {
+      throw Exception(
+          'Circular reference detected, cannot call TypeBuilderContext.jsonTypeFrom');
+    }
```
Lohann, line 148: This function checks for circular dependencies, without this, the generator runs forever.

justkawal (approved):
LGTM

PR #318, opened 2023-07-15
Fix typo in v14 metadata parser

HolyGrease:
## Description
Fix typo in metadata parser.

For example chain contains event like so:
```Rust
pub enum Event<T: Config> {
  AccountUpdated {
    account_id: T::AccountId,
    username: T::Username,
    new_username: Option<T::Username>,
    phone_number_hash: T::PhoneNumberHash,
    new_phone_number_hash: Option<T::PhoneNumberHash>,
  },
  // ...
}

```
Now after decoding chain events you get something like this, where instead of field names uses field types:
```
{
  T::AccountId: ...,
  T::Username: ...,
  // ...
}
```

But the actual value should be:
```
{
  account_id: ...,
  username: ...,
  // ...
}
```

### DIFF
--- a/packages/substrate_metadata/lib/parsers/v14_parser.dart
+++ b/packages/substrate_metadata/lib/parsers/v14_parser.dart
@@ -78,7 +78,7 @@ class V14Parser {
           // iterate over fields of each variant
           for (final v in variant['fields']) {
             args.add({
-              'name': v['typeName'],
+              'name': v['name'],
               'type': siTypes[v['type']],
             });
           }


Lohann (commented):
@HolyGrease Thank you very much for the contribution! Seems some units tests are failling, if you want to run only the unit tests of `substrate_metadata` package, you can do this:
```bash
dart run melos exec --scope="substrate_metadata"  -- 'dart test'
```

justkawal:
Closing this as no followup is taken and everything already seems to be working fine without the typo.

PR #321, opened 2023-07-19
Fix build error

### LINKED ISSUE #319: Compilation error on M1 MBP Flutter app
When including the packages from this lib in a Flutter app. Building for web, ios or android on an M1 MBP fails with the following error:
```
../../.pub-cache/hosted/pub.dev/polkadart_scale_codec-1.1.0/lib/primitives/i64.dart:11:29: Error: The integer literal 9223372036854775808 can't be represented in 64 bits.
Try using the BigInt class if you need an integer larger than 9,223,372,036,854,775,807 or less than -9,223,372,036,854,775,808.
        value > BigInt.from(9223372036854775808)) {
```

Offending code: 

https://github.com/rankanizer/polkadart/blob/main/packages/polkadart_scale_codec/lib/primitives/i64.dart
```
 @override
  void encodeTo(BigInt value, Output output) {
    if (value < BigInt.from(-9223372036854775808) ||
        value > BigInt.from(9223372036854775807)) {
      throw OutOfBoundsException();
    }
    U64Codec.codec.encodeTo(value.toUnsigned(64), output);
  }
```

Perhaps init the BigInt from a string and not from a num?
See: https://stackoverflow.com/questions/62834930/how-to-create-a-bigint-in-dart

avive:
Use parsed bigInt from string and not from num in i64 primitive to fix build error for flutter web apps.
This fixes #319 and the build error when trying to build a flutter app with polkadart_scale_codec-1.1.0:
```
../../.pub-cache/hosted/pub.dev/polkadart_scale_codec-1.1.0/lib/primitives/i64.dart:11:29: Error: The integer literal 9223372036854775808 can't be represented in 64 bits.
Try using the BigInt class if you need an integer larger than 9,223,372,036,854,775,807 or less than -9,223,372,036,854,775,808.
        value > BigInt.from(9223372036854775808)) {
```

All tests pass.

### DIFF
--- a/packages/polkadart_scale_codec/lib/primitives/i64.dart
+++ b/packages/polkadart_scale_codec/lib/primitives/i64.dart
@@ -7,8 +7,8 @@ class I64Codec with Codec<BigInt> {
 
   @override
   void encodeTo(BigInt value, Output output) {
-    if (value < BigInt.from(-9223372036854775808) ||
-        value > BigInt.from(9223372036854775807)) {
+    if (value < BigInt.parse('-9223372036854775808') ||
+        value > BigInt.parse('9223372036854775807')) {
       throw OutOfBoundsException();
     }
     U64Codec.codec.encodeTo(value.toUnsigned(64), output);


Lohann (approved):
Good catch!

Lohann:
Will do a 1.1.1 release with this fix, thanks @avive

PR #326, opened 2023-10-08
bump flutter: v3.10.1, lift dart-sdk constraint >= v3.0.1, bump http: v1.1.0

### LINKED ISSUE #325: bump http to v1.1.0
The major bump happened a few months ago, and we depend on it in our repo. Polkadart depends on 0.13.5, which creates a version conflict when I try to integrate it unfortunately.

clangenb:
This PR bumps the flutter version to v3.10.1, the earliest version that comes with the dart-sdk v.3.0.1, and also bumps http to v1.1.0, closes #325.

The http version bump was motivated because we use it downstream, but this in turn required the lift of the dart-sdk >=3.0.0.


packages/polkadart_cli/test/typegen/composite_test.dart
clangenb, line 1: The changes in this file seem like they actual fixes, where the code emitter wasn't producing properly formatted code before.

leonardocustodio (approved):

PR #328, opened 2023-10-09
Adds author_submitExtrinsic

leonardocustodio:
This PR adds author submitExtrinsic.


leonardocustodio:
@clangenb since you were interested this PR adds `author_submitExtrinsic` call. It will return to you the transaction hash.

PR #336, opened 2023-10-14
Remove lfs

clangenb:
We have to remove the lfs on the main branch of the repo, otherwise the clone will fail.


clangenb:
Sorry wrong repo

PR #337, opened 2023-10-16
[provider] fix subscriptions

### LINKED ISSUE #338: How to work with `provider.subscribe`
I could not find any tests for it, and when I tried to implement a for `author_submitAndWatchExtrinsic` I noticed that I don't get any events in my subscriptions. I also tried to get events for `chain_subscribeFinalizedHeads`.

May I ask for help please? This is the commit I am trying to fix: https://github.com/encointer/encointer-wallet-flutter/commit/1f15d2d64c152b6cda96b0c0236b2fda462aec54

clangenb:
Closes #338. So the problem is that the method name returned in subscription names doesn't match the method name that is used to actually subscribe; hence we were unable to get the correct stream controller in the subscription messages. This PR fixes it in a way that we introduce a map that uses purely the subscription ID, to identify the stream controller.

### DIFF
--- a/packages/polkadart/lib/provider.dart
+++ b/packages/polkadart/lib/provider.dart
@@ -126,9 +126,11 @@ class WsProvider extends Provider {
   /// Maps the query id to the completer that will resolve the query
   final Map<int, Completer<RpcResponse>> queries = {};
 
+  /// Maps the subscription method to the respective subscription id.
+  final Map<String, String> subscriptions = {};
+
   /// Maps the subscription id to the stream controller that will emit the subscription data
-  final Map<String, Map<String, StreamController<SubscriptionMessage>>>
-      subscriptions = {};
+  final Map<String, StreamController<SubscriptionMessage>> streamControllers = {};
 
   /// WebSocket connection
   WebSocketChannel? channel;
@@ -186,13 +188,12 @@ class WsProvider extends Provider {
     });
 
     // Subscriptions
-    jsonStream
-        .where((message) =>
-            !message.containsKey('id') &&
-            message.containsKey('params') &&
-            (message['params'] as Map<String, dynamic>)
-                .containsKey('subscription'))
-        .map((message) {
+    jsonStream.where((message) {
+      // print('Received stream msg: $message');
+      final valid =
+          !message.containsKey('id') && message.containsKey('params') && (message['params'] as Map<String, dynamic>).containsKey('subscription');
+      return valid;
+    }).map((message) {
       final method = message['method'] as String;
       final params = message['params'] as Map<String, dynamic>;
       final subscription = params['subscription'] as String;
@@ -200,9 +201,9 @@ class WsProvider extends Provider {
       return SubscriptionMessage(
           method: method, subscription: subscription, result: result);
     }).listen((message) {
-      final StreamController controller = getOrCreateSubscriptionController(
-          message.method, message.subscription);
-      controller.add(message);
+      // print('getting controller: ${message.method}, ${message.subscription}');
+      final StreamController? controller = getSubscriptionController(message.subscription);
+      controller?.add(message);
     });
 
     this.channel = channel;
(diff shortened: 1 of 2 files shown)


packages/polkadart/lib/provider.dart
```diff
@@ -186,23 +188,22 @@ class WsProvider extends Provider {
...
     // Subscriptions
-    jsonStream
-        .where((message) =>
-            !message.containsKey('id') &&
-            message.containsKey('params') &&
-            (message['params'] as Map<String, dynamic>)
-                .containsKey('subscription'))
-        .map((message) {
+    jsonStream.where((message) {
+      // print('Received stream msg: $message');
```
clangenb, line 192: I found my prints quite useful for debugging so I left them here for now. I am from the rust domain initially, and I wonder how idiomatic dart logging is done? Happy to introduce proper logging here if you advise me how you would like to do it.

packages/polkadart/lib/provider.dart
```diff
@@ -251,7 +250,13 @@ class WsProvider extends Provider {
   @override
   Future<SubscriptionReponse> subscribe(String method, List<dynamic> params,
       {FutureOr<void> Function(String subscription)? onCancel}) async {
+    // print('Subscribing to method: $method');
     final result = await send(method, params);
+
+    if (result.error != null) {
+      throw Exception(result.error.toString());
+    }
```
clangenb, line 258: Handle errors that can happen during subscribing, like subscribing in `author_submitAndWatchExtrinsic` with an invalid signature.

clangenb:
@leonardocustodio happy to receive your feedback here.

packages/polkadart/lib/provider.dart
```diff
@@ -186,23 +188,22 @@ class WsProvider extends Provider {
...
+      // print('Received stream msg: $message');
+      final valid =
+          !message.containsKey('id') && message.containsKey('params') && (message['params'] as Map<String, dynamic>).containsKey('subscription');
+      return valid;
+    }).map((message) {
       final method = message['method'] as String;
       final params = message['params'] as Map<String, dynamic>;
       final subscription = params['subscription'] as String;
-      final result = params.containsKey('result') ? message['result'] : null;
+      final result = params.containsKey('result') ? params['result'] : null;
```
clangenb, line 200: Another necessary tiny fix to make it work.

leonardocustodio:
Hey, that's great, though I had noticed this issue last week I'm still finishing up the extrinsic signing that also has an issue. I'm glad you fixed. I can review it during the day. Thanks 😄

packages/polkadart/lib/provider.dart
leonardocustodio: This package is pretty good for logging in dart, if you want to go ahead: https://pub.dev/packages/logger

leonardocustodio (commented):
The previous implementation supposed that a `subscriptionId` was unique only within each `method` but that seems to be incorrect and they are unique globally. Do you think we could simplify that and remove the `Map<String, String>` in favor of `Map<String, StreamController>` where the String is the id?

clangenb:
> The previous implementation supposed that a `subscriptionId` was unique only within each `method` but that seems to be incorrect and they are unique globally. Do you think we could simplify that and remove the `Map<String, String>` in favor of `Map<String, StreamController>` where the String is the id?

I was thinking the same thing, but then we have no way of knowing to which methods we have already subscribed to, and hence we can't prevent double subscription then. I think we want to prevent it, so that potential users of the crate can't accidentally create toons of subscriptions for the same thing.

leonardocustodio:
Although I understand that reusing the same subscription could be helpful for the developer. I'm not really sure we can say that there will be no use case where they will not want to subscribe multiple times to the same thing.
This is from `polkadot-js` https://github.com/polkadot-js/api/blob/a22a111f7228e80e03bc75298347dbd184c5630b/packages/rpc-provider/src/ws/index.ts#L533 the code is a bit confusing tbh but doesn't seem they handle that.

clangenb:
Yeah, I totally understand your reasoning, my idea was that the as long as the `params` of the subscription are the same, we'd not want to open another subscription, this would also apply to the tx, as there won't be the same tx as the nonce is increasing. 

Regardless, I am not sure yet if my logic is somewhere flawed. So I will happily change it in a way that allows for double subscription. 👍

packages/polkadart/lib/provider.dart
clangenb: Alright, I think I am a bit short on time next week. Do you mind if we just keep the print statements as comments for now and merge this PR, and we introduce proper logging in a separate PR in the future?

clangenb:
From my point of view, ready to merge!

leonardocustodio (approved):
Seems good to me, nice job 👍

PR #355, opened 2023-11-04
Create sr25519 package

justkawal:


leonardocustodio (commented):
Though the name is correct it is too difficulty to type and so to find in pub.dev let's name it sr25519

justkawal:
okay

clangenb:
May I ask for a state update on this PR?

justkawal:
> May I ask for a state update on this PR?

Its nearly completed, just adding more test-cases for more coverage....

justkawal:
Sample usage:

```dart
import 'dart:convert';
import 'package:merlin/merlin.dart' as merlin;
import 'package:sr25519/sr25519.dart';

void main() {
  final msg = utf8.encode('hello friends');
  final signingCtx = utf8.encode('example');

  final merlin.Transcript signingTranscript = Sr25519.newSigningContext(signingCtx, msg);

  final merlin.Transcript verifyTranscript = Sr25519.newSigningContext(signingCtx, msg);

  final keypair = Sr25519.generateKeyPair();

  final (priv, pub) = (keypair.secretKey, keypair.publicKey);

  final Signature sig = priv.sign(signingTranscript);

  final (verified, _) = pub.verify(sig, verifyTranscript);

  assert(verified == true);
}
```

leonardocustodio:
Isn't it integrated with the KeyPairclass yet? We shouldn't have to call sr25519 directly. Instead we just call keypair and specify the algorithm being sr25519 the default

justkawal:
> Isn't it integrated with the Keypair/Keyring class yet? We shouldn't have to call sr25519 directly. Instead we just call keyring and specify the algorithm being sr25519 the default

Yes, done the Keyring integration and with KeyPair as well....
Sorry @leonardocustodio , It took a while but was making sure everything running good with quality test-cases.

justkawal:
Example:

```dart
final keyring = Keyring(); 
// keypair creation inside keyring defaults to ed25519 but can be changed from named parameters
// with KeyPairType.sr25519

keyring.add(KeyPair.sr25519.fromSeed(your seed Uint8List value));
// or KeyPair.sr25519.fromUri(.....)
// or KeyPair.sr25519.fromMnemonic(.....)

// or replace sr25519  --->  ed25519
// or Keypair.ed25519.fromUri(.....)
// or KeyPair.ed25519.fromMnemonic(.....)

final Uint8List message = utf8.encode('this is a message');
final pair = keyring.getByPublicKey(publicKey1);
final signature = pair.sign(message);

final verified = pair.verify(message, signature); // true

// ooopps anonymous pair
final anonyPair = keyring.getByPublicKey(publicKey2);
final verified = anonyPair.verify(message, signature);  // false
```

justkawal:
While deriving seed from uri, it does that by hard and soft key derivation from the recursive junctions.

packages/sr25519/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: sr25519
+description: A starting point for sr25519 library
+version: 1.0.0
```
leonardocustodio, line 3: Though the package seems pretty complete. Let's start with a lower version like `0.1.0` so we can promote substrate_bip39 / polkadart_keyring and this all together.

packages/polkadart_keyring/lib/src/keyring.dart
```diff
@@ -9,29 +9,101 @@ part of polkadart_keyring;
 /// Example usages for each function are provided below.
 class Keyring {
   final Pairs pairs;
+  final KeyPairType keyPairType;
 
-  // The default SS58 address format used by the keyring.
-  int _ss58Format = 42;
+  Keyring()
+      : pairs = Pairs(),
+        keyPairType = KeyPairType.ed25519;
```
leonardocustodio, line 16: As the default on-chain as `sr25519` our default should also be `sr25519`

packages/polkadart_keyring/lib/src/keypair.dart
```diff
@@ -34,142 +58,170 @@ class KeyPair {
...
+  String get address;
+
+  /// Get the public key.
+  ///
+  /// Example:
+  /// ```dart
+  /// final keyPair = KeyPair.sr25519.fromSeed(seed); // Replace with your actual seed
+  /// print('Public Key: ${keyPair.publicKey}');
+  /// ```
+  dynamic get publicKey;
```
leonardocustodio, line 182: What type is this? On ed25519 we can do:
`keypair.publicKey.bytes` how we do the same on `sr25519`? We should have the same interfaces.

packages/sr25519/CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## 1.0.0
```
leonardocustodio, line 1: Don't forget to lower here

leonardocustodio (changes requested):
Overall the PR looks real good. Awesome job mate.

packages/sr25519/CHANGELOG.md
justkawal: Resolved

packages/polkadart_keyring/lib/src/keypair.dart
justkawal: Resolved

justkawal:
Resolved conflicts and updated @leonardocustodio 
Waiting for review....

packages/polkadart_keyring/lib/src/sr25519.dart
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @override
+  String get address {
+    return Address(prefix: ss58Format, pubkey: bytes).encode();
+  }
+
+  @override
+  Future<void> unlockFromMemonic(String mnemonic, [String? password]) async {
+    final seed = await SubstrateBip39.ed25519.seedFromUri(mnemonic);
```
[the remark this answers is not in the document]
justkawal: Actually this is right, because in rust bindings they are doing the same.
Analysed and converted from the rust code directly, so reused the functionality.

But also I'm open to see the `test-cases` for which It will fail....

packages/polkadart_keyring/lib/src/ed25519.dart
```diff
@@ -0,0 +1,97 @@
...
+    return ed.verify(_publicKey, message, signature);
+  }
+
+  @override
+  String get address {
+    return Address(prefix: ss58Format, pubkey: bytes).encode();
+  }
+
+  @override
+  Future<void> unlockFromMemonic(String mnemonic, [String? password]) async {
```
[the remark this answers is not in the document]
justkawal: Resolved

packages/polkadart_keyring/lib/src/keypair.dart
```diff
@@ -34,142 +58,170 @@ class KeyPair {
...
   /// keyPair.unlockFromMemonic(mnemonic);
   /// keyPair.sign(message); // Works
   /// ```
-  Future<void> unlockFromMemonic(String mnemonic) async {
-    final seed = await SubstrateBip39.ed25519.seedFromUri(mnemonic);
-    _unlock(ed.newKeyFromSeed(Uint8List.fromList(seed)));
-  }
+  ///
+  ///
+  Future<void> unlockFromMemonic(String mnemonic, [String? password]);
```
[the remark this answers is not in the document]
justkawal: Resolved

leonardocustodio:
Can you rebase the pr? It is showing more changes than it should

packages/polkadart_keyring/lib/src/sr25519.dart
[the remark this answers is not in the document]
clangenb: Alright, I don't have the technical insight here, but the naming just put me off. In this case, a comment might be nice that it is actually correct. :)

leonardocustodio:
There is only a single issue. At the `ExtrinsicPayload` class. You can see there is the `signatureType` it was hardcoded to `ed25519` as that was the only type we supported. Now, we need to set it to `00` if `ed25519` or `01` to `sr25519`. 

In this case I suggest to also make an extension where you can pass the keypair to an `Extrinsic` like it is done in polkadot-js as the following:

```
	const { signature } = registry
		.createType('ExtrinsicPayload', signingPayload, {
			version: EXTRINSIC_VERSION,
		})
		.sign(pair);
````

Would also be nice to have a helper function where I could sign the SigningPayload (not the Extrinsic class) like so:
```
	const signature = signWith(alice, signingPayload, {
		metadataRpc,
		registry,
	});
```

Besides that, everything looks really good, congratulations my friend.

**Comment by justkawal** [comment]
Can you provide the permalink to the ExtrinsicPayload class?

> There is only a single issue. At the `ExtrinsicPayload` class. You can see there is the `signatureType` it was hardcoded to `ed25519` as that was the only type we supported. Now, we need to set it to `00` if `ed25519` or `01` to `sr25519`.
> 
> In this case I suggest to also make an extension where you can pass the keypair to an `Extrinsic` like it is done in polkadot-js as the following:
> 
> ```js
> 	const { signature } = registry
> 		.createType('ExtrinsicPayload', signingPayload, {
> 			version: EXTRINSIC_VERSION,
> 		})
> 		.sign(pair);
> ```
> 
> So for us, maybe something like this:
> 
> ```dart
>   final extrinsic = Extrinsic(
>     signer: publicKey,
>     method: encodedCall,
>     signature: hexSignature,
>     eraPeriod: 64,
>     blockNumber: blockNumber,
>     nonce: 0,
>     tip: 0,
>   ).encode(api.registry).sign(pair);
> ```
> 
> Would also be nice to have a helper function where I could sign the SigningPayload (not the Extrinsic class) like the following example of `tx-wrapper`:
> 
> ```js
> 	const signature = signWith(alice, signingPayload, {
> 		metadataRpc,
> 		registry,
> 	});
> ```
> 
> Besides that, everything looks really good, congratulations my friend.

**Comment by leonardocustodio** [comment]
> Can you provide the permalink to the ExtrinsicPayload class?

https://github.com/leonardocustodio/polkadart/tree/main/packages/polkadart/lib/extrinsic

File: packages/polkadart_keyring/pubspec.yaml [hunk 8]
```diff
@@ -16,6 +17,10 @@ dependencies:
   typed_data: ^1.3.1 # BSD-3-Clause
   ed25519_edwards: ^0.3.1
   convert: ^3.1.1
+  sr25519:
+    path: ../sr25519
```
clangenb, line 21: Think that dart doesn't like this. The Pubspec.lock refers to my local filesystem, and I can't fix it with dependency overrides.

clangenb: This seems to be a problem with git deps only: https://github.com/dart-lang/pub/issues/3336

clangenb: So it seems to be a chicken-and-egg problem. May I ask you you to publish the sr25519 package to pub.dev, and then adjust the dependency here to refer to pub.dev. With this, I suppose that I can safely rely on my fork of polkadart again without any issues.

leonardocustodio: Yep, it needs to be published first to change it as the package doesn't exists yet. We will do as soon as the remaining comments are addressed

justkawal:
@leonardocustodio 

Updated and made a followup PR: #386

leonardocustodio:
Thanks, I'm back from holidays, I will be reviewing this in the end of the day.

leonardocustodio (approved):

leonardocustodio:
sr25519 published, feel free to make the changes on pubspec dependencies @justkawal 
https://pub.dev/packages/sr25519
Btw great job!!!

PR #371, opened 2023-11-20
[provider] fix `isConnected` check for websocket provider

clangenb:
`isConnected` erroneously returned false because the check of the close code had the wrong operation.


leonardocustodio (approved):
lgtm, thank you for your contribution @clangenb

leonardocustodio:
Version 0.2.3 released with the fix https://pub.dev/packages/polkadart

PR #381, opened 2023-12-01
Generate storage key prefix methods

### LINKED ISSUE #374: Expose (double) map prefix for generated storage types or expose a `keys()` method for (double) maps
It would be awesome to have that to easily generate storage keys for the `state_getKeysPaged` method. I imagine something like:

```dart
  Uint8List mapKeyPrefix() {
    final Uint8List hash = Uint8List(32 + hasher.size(key));
    Hasher.twoxx128.hashTo(
        data: Uint8List.fromList(utf8.encode(prefix)),
        output: hash.buffer.asUint8List(hash.offsetInBytes, 16));
    Hasher.twoxx128.hashTo(
        data: Uint8List.fromList(utf8.encode(storage)),
        output: hash.buffer.asUint8List(hash.offsetInBytes + 16, 16));
    return hash;
  }
```

Additionally, the generated storage types must be make public, to easily access that. I imagine something like:

```dart
final prefix = polkadot.query.system.account.mapKeyPrefix();
final keys = polkadot.rpc.state.getKeysPaged(prefix, count: 50);
```

Or this could even be internalized as generated code:
```dart
final keys = polkadot.query.system.account.keys(count: 50);
```

clangenb:
The generated code here can be used in `polkadot.rpc.state.getKeysPaged` RPC, this code has already been integrated into our app codebase, and it works. Proof of concept in our [app](https://github.com/encointer/encointer-wallet-flutter/blob/7f774f8d3796b9ce81850e65744364b7093e55be/app/lib/service/substrate_api/encointer/encointer_api.dart#L633-L658). There is also a new examples in `examples/bin/get_account_infos_demo.dart`

* Closes #374


packages/polkadart/lib/substrate/storage.dart
```diff
@@ -115,15 +115,25 @@ class StorageMap<K, V> {
 
   Uint8List hashedKeyFor(K key) {
     final Uint8List hash = Uint8List(32 + hasher.size(key));
+    _hashPrefixTo(output: hash);
```
clangenb, line 118: Extracted this piece of code to prevent redundancy.

packages/polkadart_cli/lib/src/generator/pallet.dart
```diff
@@ -448,13 +448,9 @@ Class createPalletQueries(
...
-                  ..statements.add(declareFinal('hashedKey')
-                      .assign(refer('_$storageName')
-                          .property(storage.hashers.isEmpty
-                              ? 'hashedKey'
-                              : 'hashedKeyFor')
-                          .call(storage.hashers.map((hasher) => refer(
-                              'key${storage.hashers.indexOf(hasher) + 1}'))))
+                  ..statements
+                      .add(declareFinal('hashedKey')
+                      .assignHashedKey(storageName, storage)
```
clangenb, line 453: Minor refactoring to prevent duplicate code.

packages/polkadart_cli/lib/src/generator/pallet.dart
```diff
@@ -583,7 +596,50 @@ Class createPalletConstants(
...
+extension MethodNameExtension on Storage {
+
+  /// Name of the generated method returning the key of a storage.
+  String keyMethodName() {
+    return '${name}Key';
+  }
+
+  /// Name of the generated method returning the key prefix of a storage.
+  String mapPrefixMethodName() {
+    return '${name}MapPrefix';
```
clangenb, line 608: I have been pondering a lot about this name, classical trade-off between verbosity and expressiveness. Let me now, if you disagree with the naming.

packages/polkadart_cli/lib/src/generator/pallet.dart
```diff
@@ -583,7 +596,50 @@ Class createPalletConstants(
...
+      throw Exception(
+        'Bad code generation path; can\'t create keyPrefix method without hashers.',
+      );
+    }
+
+    if (storage.hashers.length > 2) {
+      throw Exception(
+        'Bad code generation path; keyPrefix method for maps with depth > 2 are not supported yet.',
+      );
+    }
```
clangenb, line 633: Before implementing it for all, I wanted to get ensure my path is on aligned with your ideas. Also, 99% of the maps are probably not deeper than 2.

clangenb:
Anything missing from my side to get a review here? :)

leonardocustodio:
I will do it this one and #355 tomorrow, sorry for the delay.

justkawal:
> I will do it this one and #355 tomorrow, sorry for the delay.

It's alright, no worries.

packages/polkadart_cli/lib/src/generator/pallet.dart
leonardocustodio: The PR is really good and it will be very useful for querying storage. Just like you I just wonder about the name.
The RPC query (getKeysPaged) takes a `PrefixedStorageKey` or `StorageKey`.

Though I understand that `MapPrefix`means that you are getting whatever is behind and asking it prefix. I wonder if it would not make more sense to instead say the action it will do just say what it is.

Example: accountStorageKey or accountPrefixedKey

I mean if we take for example: "ABC".toHex() it says the action but we could have anything in that string. While these auto-generated classes even though I can have another thing instead of `account` as the above. It is already called the "specific" data. I mean, it is static, it doesn't need to be parsed like "ABC".toHex() would.

So instead I believe just saying what will come in return sounds a bit more clear than saying what it will do. No sure If I explained well my reasoning.

But again, I believe both views are right. I would like to know @Lohann thoughts on this one too.

packages/polkadart/lib/substrate/storage.dart
leonardocustodio: Sounds good

leonardocustodio:
If you don't mind would you be able to put the example here as well: https://github.com/leonardocustodio/polkadart/blob/main/examples/bin/demo.dart

clangenb:
Thanks for the feedback! I will look at this on wednesday.

packages/polkadart_cli/lib/src/generator/pallet.dart
clangenb: I think in general we align on this topic, my idea was also to name it according to what it is: the storage map prefix, and hence, to take your example, call it `accountMapPrefix` because it is the storage map prefix of the account storage.

To me, it is more intuitive like that because reading `accountPrefixedKey` reads to me like: it is the full key to an account in the storage including the prefix, which is not true. However, if it is called `accountMapPrefix`, it is clear (to me at least) that is only the storage map prefix and not the full key.

I was also considering to include the depth of the map into the name, which would result in `accountDoubleMapPrefix`, but I considered that an overkill.

leonardocustodio: Well, I don't see any problems in the PR, and those details can always be changed another time if needed. Let's not delay this and merge. Were you able to add the example?

examples/lib/README.md
```diff
@@ -0,0 +1 @@
+Directory for the generated files.
```
clangenb, line 1: This directory has to exist to be able to generate the types, I thought there is no harm in committing that, and then we can follow 1:1 the README guidance.

melos.yaml
```diff
@@ -7,6 +7,7 @@ command:
 
 packages:
   - packages/**
+  - examples
```
clangenb, line 10: To make my life easier, I added the examples project to melos. Otherwise, I can't refer to the local packages, as it tries to get the ones from pub.dev.

examples/analysis_options.yaml
```diff
@@ -13,18 +13,8 @@
...
-# For more information about the core and recommended set of lints, see
-# https://dart.dev/go/core-lints
-
-# For additional information about configuring this file, see
-# https://dart.dev/guides/language/analysis-options
+linter:
+  rules:
+    # Generated files don't satisfy this.
+    avoid_renaming_method_parameters: false
+    no_leading_underscores_for_local_identifiers: false
```
clangenb, line 20: Added linter rules such that we don't get warnings after generating the types.

examples/bin/extrinsic_demo.dart
```diff
@@ -1,10 +1,11 @@
 import 'package:convert/convert.dart';
-import '../lib/generated/polkadot/polkadot.dart';
-import '../lib/generated/polkadot/types/sp_runtime/multiaddress/multi_address.dart';
 import 'package:polkadart/polkadart.dart'
     show AuthorApi, Extrinsic, Provider, SigningPayload, StateApi;
 import 'package:polkadart_keyring/polkadart_keyring.dart';
 
+import 'package:polkadart_example/generated/polkadot/polkadot.dart';
+import 'package:polkadart_example/generated/polkadot/types/sp_runtime/multiaddress/multi_address.dart';
```
clangenb, line 7: I got a linter warning for relative packages imports to I changed this.

examples/pubspec.yaml
```diff
@@ -17,6 +17,9 @@ dependencies:
   substrate_bip39: ^0.2.0
   substrate_metadata: ^1.1.0
 
+  # Generated files depend on quiver
+  quiver: ^3.2.1
```
clangenb, line 21: This was also a linter warning.

examples/bin/extrinsic_demo.dart
clangenb, line 1: I hope you align with my decision here to do two demos that serve a single purpose.

clangenb:
I have added the example, ready to review from my point of view.

leonardocustodio (approved):
lgtm, thanks for the contribution one more time 😃

leonardocustodio:
Released under:
```
polkadart: ^0.3.0
polkadart_cli: ^0.3.0
```

PR #382, opened 2023-12-19
[CI] fail if fmt is wrong and check `polkadart_example`

clangenb:
This PR introduces the following changes:
* The fmt CI command didn't return an error if there were badly formatted fails, this is changed now.
* Check the code of the `polkadart_example` in the CI. This means that we also generate the types for Polkadot in the CI.


leonardocustodio (approved):
Really good job, thanks!

Issue #359, opened 2023-11-07
Custom Signed Extensions


justkawal:

clangenb:
I see you have assigned this to yourself, may I ask when you intend to implement this?

leonardocustodio:
Id everything goes as planned I was hoping to finish this by the end of the month

clangenb:
As usual, thanks a lot for your quick reply. This helps a lot with planing! 😊

leonardocustodio:
I've unsigned myself as @justkawal has already made some progress on this, we should have the PR really soon.

leonardocustodio:
Released under:
```
polkadart: ^0.4.0
polkadart_scale_codec: ^1.2.0
substrate_metadata: ^1.2.0
```

PR #401, opened 2024-01-26
[extrinsic] fix extrinsic codec for `AssetHubExtensions`

clangenb:
I am implementing in a naive way our custom signed extensions, and while doing that I noticed that the asset hub signed extensions didn't work. This PR fixes that. I am doing this in the quick way because I know that you are implementing https://github.com/leonardocustodio/polkadart/issues/359, which will probably change this part of the code anyhow.

### DIFF
--- a/packages/polkadart/lib/extrinsic/extrinsic_payload.dart
+++ b/packages/polkadart/lib/extrinsic/extrinsic_payload.dart
@@ -30,7 +30,7 @@ class Extrinsic {
     this.assetId,
   });
 
-  toEncodedMap() {
+  toEncodedMap(dynamic registry) {
     return {
       'signer': signer,
       'method': method,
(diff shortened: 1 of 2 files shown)


packages/polkadart/lib/extrinsic/extrinsic_payload.dart
```diff
@@ -68,7 +68,8 @@ class Extrinsic {
     final List<String> extras = <String>[];
 
     late final SignedExtensions signedExtensions;
-    if (assetId != null) {
+
+    if (_usesChargeAssetTxPayment(registry)) {
```
clangenb, line 72: This was wrong, even if the `assetId` is null, the format is different, namely the `None` needs to be added.

packages/polkadart/lib/extrinsic/extrinsic_payload.dart
```diff
@@ -108,4 +109,16 @@ class Extrinsic {
...
     return keyPair.sign(encode(registry, signatureType));
   }
+
+  bool _usesChargeAssetTxPayment(dynamic registry) {
+    return registry.getSignedExtensionTypes().contains('ChargeAssetTxPayment');
+  }
+
+  String maybeAssetIdEncoded(dynamic registry) {
+    if (_usesChargeAssetTxPayment(registry)) {
+      return assetId != null ? assetId!.toRadixString(16) : '00';
```
clangenb, line 119: Added the `00` in case the assetId is null.

packages/polkadart/lib/extrinsic/extrinsic_payload.dart
leonardocustodio: Nice catch, thanks

leonardocustodio (approved):

PR #402, opened 2024-01-26
fix paying extrinsic with custom asset

clangenb:
My testing for #401 was unfortunately not good enough; it only worked when the `assetId` was null when sending extrinsics. This PR fixes now the non-null case.


leonardocustodio:
Released under `polkadart: 0.3.2`

PR #446, opened 2024-03-22
Add types 'CheckAppId' for Avail chain

gtmickey:
Avail has more Registry types, add CheckAppId for Payload encode.


gtmickey:
Can be done with customSignedExtensions, closed.

PR #460, opened 2024-06-21
Add support for `CheckMetadataHash` extension

clangenb:
This doesn't implement the full feature of the signed extensions. However, signing the actual metadata hash is only valuable for offline clients.

This change makes the api at least compatible with the fellows/runtimes version v1.2.5, i.e. the current versions of Kusama/Westend, and it will also ensure that it works for Polkadot/Paseo after the next runtime upgrade.


leonardocustodio (approved):

leonardocustodio:
Thank you for your contribution, the changes have been released under: `polkadart 0.4.5`

PR #463, opened 2024-06-27
Add missing error value parse for http provider

weishirongzhen:
When send a rpc from http provider if there is an error, should set error value to RpcResponse, or else there is no way to get exception message by user.


leonardocustodio (approved):

leonardocustodio:
Thank you for the contribution. Released under `polkadart 0.4.6`

Issue #504, opened 2025-01-09
ChargeAssetTxPayment fails with non int asset id


clangenb:
Minimal example: https://github.com/encointer/polkadart/commit/b03dfa1ead471c1e3e3a5da70066779b8616d52a

The reason is that proxying the codec via the `toJson` produces a wrong output: https://github.com/leonardocustodio/polkadart/blob/b03dfa1ead471c1e3e3a5da70066779b8616d52a/packages/polkadart/lib/primitives/community_identifier.dart#L30

We need to use the array codec, but when we go via the json, it is presumably encoded as a vec, which means that we have an extra byte in the codec, which is wrong.

clangenb:
The solution would probably be to allow passing the object directly as the asset id, but then I currently get the error:

```console
Encoding {tip: 0, asset_id: Instance of 'CommunityIdentifier'}
Unhandled exception:
type 'CommunityIdentifier' is not a subtype of type 'Map<String, dynamic>' of 'value'
#0      CompositeCodec.encodeTo (package:polkadart_scale_codec/primitives/composite.dart:27:38)
#1      NestedOptionCodec.encodeTo (package:polkadart_scale_codec/primitives/option.dart:16:20)
#2      CompositeCodec.encodeTo (package:polkadart_scale_codec/primitives/composite.dart:39:13)
#3      ExtrinsicPayload.encode (package:polkadart/extrinsic/extrinsic_payload.dart:150:16)
#4      main (file:///C:/code/polkadart/examples/bin/extrinsic_demo.dart:106:7)
```

Because we call `encodeTo` on the option, which expects a json as an argument even though the `encodeTo` from the community identifier would correctly encode on the object itself: https://github.com/leonardocustodio/polkadart/blob/b03dfa1ead471c1e3e3a5da70066779b8616d52a/packages/polkadart/lib/primitives/community_identifier.dart#L61

clangenb:
Nevermind, I forgot to add the signed extensions to the signingPayload, see my fix: https://github.com/encointer/polkadart/commit/b6a274c5ef5a71327d0face9f81e225ad767600d.

The error was on my end and everything works flawlessly! Great work with the signed extension!

clangenb:
Here would be the full working example of using a `ChargeAssetTxExtension` payment with a complex asset id type:

* https://github.com/leonardocustodio/polkadart/compare/main...encointer:polkadart:cl/encointer-signed-ext-updated

leonardocustodio:
[...]

PR #506, opened 2025-01-09
Add an example for using a custom `ChargeAssetTxPayment` with the encointer chain

clangenb:
As per https://github.com/leonardocustodio/polkadart/issues/504, we thought it makes sense to have an example showing how to actually use a custom signed extension.


leonardocustodio:
Awesome, thank you very much for the contribution once again

leonardocustodio (approved):

Issue #663, opened 2025-07-28
polkadart_cli generates weird character in referenda.dart file


quetool:
`TrackDetails`'s name is being generated with weird character and this of course fails to compile.

<img width="437" height="100" alt="Image" src="https://github.com/user-attachments/assets/efd8d951-efe7-4712-9ab2-3748542ff8dc" />

I'm running latest version of the cli

```
Alfredos-MacBook-Pro:reown_appkit alfreedom$ flutter doctor -v
[✓] Flutter (Channel stable, 3.32.8, on macOS 15.4.1 24E263 darwin-arm64, locale en-US) [546ms]
    • Flutter version 3.32.8 on channel stable at /Users/alfreedom/Development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision edada7c56e (3 days ago), 2025-07-25 14:08:03 +0000
    • Engine revision ef0cd00091
    • Dart version 3.8.1
    • DevTools version 2.45.1

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [2.3s]
    • Android SDK at /Users/alfreedom/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • ANDROID_HOME = /Users/alfreedom/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.5+-12932927-b750.29)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [1,132ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16F6
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [11ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.3) [11ms]
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
[...]

PR #664, opened 2025-07-28
Fixes StrCodec where null bytes are not stripped on fixed size arrays

leonardocustodio:
This should fix the following issue https://github.com/leonardocustodio/polkadart/issues/663

**Note: I'm not convinced this is correct. I suppose that maybe the wrong size is being returned? Why would we have a string array bigger than the string itself?**

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Strip trailing null bytes in StrCodec.decode

- Add direct tests for StrCodec null handling

- Bump package version to 1.5.1

- Add local dependency overrides across packages

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Input bytes"] --> B["utf8.decode(bytes)"]
  B --> C["Find null byte index"]
  C -- "no null" --> D["Return full string"]
  C -- "null found" --> E["Return substring before null"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody></tr></tbody></table>

</details>

___

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Contributions of any kind are always welcome! ❤️
 
 ## Contributors
 
-A heartfelt thank you to all the contributors for enriching the Polkadart project! [emojis](https://allcontributors.org/docs/en/emoji-key)
+A heartfelt thank you to all the contributors for enriching the Polkadart project! [emojis](https://github.com/all-contributors/all-contributors/blob/master/docs/emoji-key.md)
 
 <!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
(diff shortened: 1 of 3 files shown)


leonardocustodio:
@Lohann can you comment on this one? Should we worry about any edge cases?

leonardocustodio:
I'm going to merge this one for now to unblock https://github.com/leonardocustodio/polkadart/issues/663 but this needs to be reviewed to find out why this is happening. I will only merge as all tests passes and this doesn't not seen to be breaking anything.

PR #691, opened 2025-09-13
Fix: Entropy property on newer library version

justkawal:


leonardocustodio (approved):

leonardocustodio:
Can you also bump the package version and add it to the changelog?

justkawal:
Sure

